PR #59, opened 2015-03-09
Fix compile error in linux

qjia7:
This change will remove the var which is set but never used.


qjia7:
Ping @jeanlf @rbouqueau

jeanlf:
fixed by adding a log, thanks for the report

PR #64, opened 2015-03-16
Fix typo in DASH-AVC/264 profiles.

tieleman:
The DASH-AVC/264 settings contain an invalid profile ("`guildelines`" vs. "`guidelines`"). This causes it to not be compliant to the spec.


jeanlf:
Thanks ! Now merged

PR #90, opened 2015-04-02
Always send duration info.

canatella:
For some live HLS streams, duration event are not send. It's interesting to always get this event as one can then wait for it before displaying a progress bar for example. 

It's not sent because of a return in gf_scene_set_duration() that prevent the event to be sent.


jeanlf:
the event is not sent if the duration is the same, but you should have the first event notifying this duration no ?

canatella:
Nope, if I play a live stream, duration and max duration are 0, I never get
a notification so I do not know if the stream is seekable or not.

jeanlf:
OK I see your pb but that won't help much: if the event is not send this means the scene is not seekable. For timeshifting of live there are events defined but these are not dispatched outside of the player yet, wouldn't that be preferable ?

canatella:
Well it's working for our live stream as is but indeed if you have the
right events, I'd prefer to get those. The thing is, as is, I cannot tell
if a stream is seekable or not, because not receiving it at one point does
not mean I won't receive it later. Or is there an event after which I can
say "if I have not received a duration event now, it means its not
seekable".

PR #91, opened 2015-04-02
[Android] Do not call glDisable when it is not needed.

canatella:
In the Android video out driver, glDisable is called at init and then
for each rendering step. As per the glEnable man pages, the capability
flags that are disabled in the driver are disabled by default:

    The initial value for each capability with the exception of
    GL_DITHER and GL_MULTISAMPLE is GL_FALSE.

So there is no need to call glDisable for those. Besides, it causes
spurious log messages on some Android devices:

    W/Adreno-ES20(24368): <process_gl_state_enables:519>: GL_INVALID_ENUM

which slows down the rendering process.


jeanlf:
indeed thanks for the patch

Issue #110, opened 2015-04-17
[DASH/HLS] Crash when disconnecting while dash thread is running


canatella:
On PC, start a dash or HLS stream and send q to the terminal before the service connected message appears, it will trigger a crash in the dash thread as soon as the service is connected. On unix, for example:
 
    echo q | /Applications/Osmo4.app/Contents/MacOS/Osmo4 http://dash.edgesuite.net/dash264/TestCases/1a/netflix/exMPD_BIP_TC1.mpd 

It's a problem because on android, the application can be paused at any moment.

jeanlf:
Did #111 fixed this ?

rbouqueau:
Did #111 fixed this ?

PR #111, opened 2015-04-17
Fix dash cleanup path.

canatella:
When closing the dash thread, make sure that it is stopped by checking
for GF_DASH_STAE_STOPPED. Checking for not GF_DASH_STATE_RUNNING is not
sufficient if the thread is being stopped before the main loop of the
thread.

This might be related to issue #110 but there are also other problem in there.

I also removed the reset of dash->mpd_stop_request in the thread as it creates a race condition if the thread is closed before it.


jeanlf:
thanks make sense indeed

PR #112, opened 2015-04-17
Fix building without VRML.

canatella:
I suppose the VRML tag definitions are needed for mpeg4 stuff. I did not completely removed the ifdefs, if your ok with this, I'll remove them.


jeanlf:
Not sure what you are trying to fix here, could you give more details ?

jeanlf:
should be integrated now

PR #113, opened 2015-04-17
Cleanup android audio driver

canatella:
This is a cleanup of the android audio driver. I made it use the GPAC log facility. 
Refactored a bit the JNI functions and made sure that no static storage is used for variables relating to the driver context. 

It fixes crashes when pausing/resuming the Android player activity because of not correctly attached or detached JNI threads. It also fixes the cleanup path when releasing resources.


fcsteagu:
the gm_droid_audio.mk (build/jni/modules) needs also an update otherwise the compiling chain crashes in there :)

PR #126, opened 2015-04-27
Patch for CENC, IV was being skipping +1 bytes incorrectly

nwind21:
IV in CENC was being skipped +1 bytes incorrectly in between NAL units.  This causes a IV desynchronization after the first counter is used since every IV following is skipping 1 byte.

### DIFF
--- a/src/mcrypt/ctr.c
+++ b/src/mcrypt/ctr.c
@@ -93,14 +93,20 @@ GF_Err _mcrypt_set_state(void *_buf, void *IV, int size)
 
 GF_Err _mcrypt_get_state(void *buf, void *IV, int *size)
 {
-	if (*size < ((CTR_BUFFER* )buf)->blocksize + 1) {
-		*size = ((CTR_BUFFER* )buf)->blocksize + 1;
+	if (*size < ((CTR_BUFFER* )buf)->blocksize) { 
+		*size = ((CTR_BUFFER* )buf)->blocksize;
 		return GF_BAD_PARAM;
 	}
-	*size = ((CTR_BUFFER* )buf)->blocksize + 1;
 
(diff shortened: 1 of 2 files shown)


src/mcrypt/ctr.c
```diff
@@ -93,14 +93,20 @@ GF_Err _mcrypt_set_state(void *_buf, void *IV, int size)
 
 GF_Err _mcrypt_get_state(void *buf, void *IV, int *size)
 {
-	if (*size < ((CTR_BUFFER* )buf)->blocksize + 1) {
-		*size = ((CTR_BUFFER* )buf)->blocksize + 1;
+	if (*size < ((CTR_BUFFER* )buf)->blocksize) { 
+		*size = ((CTR_BUFFER* )buf)->blocksize;
 		return GF_BAD_PARAM;
 	}
-	*size = ((CTR_BUFFER* )buf)->blocksize + 1;
```
nguyen-viet-thanh-trung: For ctr, size in _mcrypt_get_state (and also _mcrypt_set_state) should be blocksize + 1: first byte is the current counter position; following by blocksize of IV

nwind21: Can you point me to the location in the CENC spec where the it should be blocksize+1 for the current counter after a subsample encrypt (with remaining counter)?  I didn't see anything that would suggest we needed to +1 byte but we may have missed something.

Our CENC decryption routines don't skip any bytes.  E.g. if 8 bytes of counter is left, we consume the 8 bytes of the counter.

nguyen-viet-thanh-trung: Perhaps the variable name in getter/setter routines make a confusion : the IV in set state and get state is in fact the current counter position + IV. In setter is copy first byte to c_counter_pos and to c_counter (i.e the counter which is used in encryption/decryption) so any byte was skipped.

jeanlf:
please see #135

PR #142, opened 2015-05-11
Dash avg duration

rbouqueau:
The aim of this branch is to allow a more convenient use of MP4Box with dash-ctx. Currently it works great when fed with HLS-like content. However some users want to use it to generate a live-stream from a static file which required:
1) Align the duration across tracks (':dur=' option);
2) Have the mean duration of segments respect the input of the user (i.e. do not let segment number drift across representations of adaptation sets). It would help us gain compatibility since Wowza does this and some popular players (such as VisualOn) do not react well with our streams.

When the pull request is done, I'll write a blog article.

Reviews welcomed @gpac/contributors


rbouqueau:
@jeanlf  I'd really like to merge this. The only issue is d153b4cd3006506c8e2d63807cde57b20f81c72d that could deserve an option or a version number bump ;) (since MP4Box will change its behaviour when DASHing). Opinion?

Issue #147, opened 2015-05-17
GPAC API: assert when removing sample


rbouqueau:
Hi,

When removing a sample, I sometimes (depending on the content) get a raised assert at:
```
//always called before removing the sample from SampleSize
GF_Err stbl_RemoveCTS(GF_SampleTableBox *stbl, u32 sampleNumber)
{
	GF_CompositionOffsetBox *ctts = stbl->CompositionOffset;
	assert(ctts->unpack_mode); //ASSERT
	[...]
```

Content to trigger: http://download.tsi.telecom-paristech.fr/gpac/DASH_CONFORMANCE/TelecomParisTech/sources/counter_video_360.mp4

Callstack:
```
>	libgpac.dll!stbl_RemoveCTS(GF_SampleTableBox * stbl, unsigned int sampleNumber)  Line 934	C
 	libgpac.dll!gf_isom_remove_sample(__tag_isom * movie, unsigned int trackNumber, unsigned int sampleNumber)  Line 1040 + 0x1a bytes	C
 	__caller__

```

The removal code:
```
{
	u32 di;
	GF_ISOSample *s = gf_isom_get_sample(in, trackNumber, i, &di);
	e = gf_isom_remove_sample(in, trackNumber, i); //CRASH
	gf_isom_sample_del(&s);
	assert(e == GF_OK);
}
```

Ideas?

rbouqueau:
Related to #142

PR #148, opened 2015-05-18
fix bug in ctts unpack mode when removing sample

nguyen-viet-thanh-trung:
I think that we need unpack also CompositionOffsetBox before editing/removing the entries (cf. #147 )


rbouqueau:
Excellent, thank you :)

Issue #41, opened 2015-01-25
[DASH] Add missing DASH-IF IOP profiles.


vladimir-kazakov:
Hi!

While creating DASH HEVC media according to DASH-IF IOP guidelines (I'm using the publicly available version v2.9, which is a draft of soon to be released v3.0 (http://dashif.org/w/2014/08/DASH-IF-IOP-v2.90.pdf), I noticed that MP4Box knows only about DASH and DASH-264/AVC profiles (at least that's what I see in `mp4box -h dash` (`-profile` option)). Since HEVC is now also the part of DASH-IF IOP, it would be nice to have additional profiles for it. Perhaps, something similar to DASH-264/AVC, for example: "dashevc265:onDemand" and "dashevc265:live".

cconcolato:
Thanks. It will be added once the DASH spec is stable.

vladimir-kazakov:
DASH-IF IOP v3.0 is released (http://dashif.org/wp-content/uploads/2015/04/DASH-IF-IOP-v3.0.pdf).

So, instead of adding HEVC specific profiles (they're actually called "interoperability points", but for the sake of simplicity, I'll refer to them as "profiles"), they defined new profiles. And some of these profiles cover both AVC and HEVC. So, I'm going to rename this issue, because now both AVC and HEVC profiles are missing.

But while figuring out what have to be done, I actually noticed that MP4Box currently works incorrectly regarding profiles.

The problem is that currently users can specify 2 profiles: `dashavc264:onDemand` and `dashavc264:live`. The part after the colon specifies the MPEG-DASH profile, so the `urn:mpeg:dash:profile:isoff-on-demand:2011` identifier will be used in the ":onDemand" case, and `urn:mpeg:dash:profile:isoff-live:2011` in the ":live" case. And the part before the colon tells to MP4Box that this is the DASH-IF profile, so it adds the `http://dashif.org/guidelines/dash264` identifier to the list of profiles. This is correct, but not always. According to the description of this DASH-IF IOP profile (section 6 in any version of DASH-IF IOP), it may only be used for media up to 720p. So, it should not be possible to create 1080p media, at least without any warning. But now, of course, it's possible, which is wrong. There's a way to add custom identifiers by using the `-profile-ext` parameter, but users still can't override this "dash264" identifier that MP4Box always uses. I'm not going to create a separate issue for this bug, because I think that it will disappear after the current issue will be fixed.

So, here is the list of problems that I see:

[...]

PR #288, opened 2015-10-19
DASH-IF IOP profiles require ContentProtection only in AdaptationSet.

vladimir-kazakov:
DASH-IF IOP 3.1 says in 7.5.2 that:

> ContentProtection Descriptors SHALL always be present in the AdaptationSet element, and apply to all contained Representations

The same is related to previous IOP versions. ContentProtection was never allowed to be in Representation.

### DIFF
--- a/applications/mp4box/main.c
+++ b/applications/mp4box/main.c
@@ -386,6 +386,8 @@ void PrintDASHUsage()
 	        " -single-traf         uses a single track fragment per moof (smooth streaming and derived specs may require this)\n"
 	        " -dash-ts-prog N      program_number to be considered in case of an MPTS input file.\n"
 	        " -frag-rt             when using fragments in live mode, flush fragments according to their timing (only supported with a single input).\n"
+	        " -cp-in-as            if media is protected, adds the ContentProtection element to the AdaptationSet element.\n"
+	        " -cp-in-rep           if media is protected, adds the ContentProtection element to the Representation element.\n"
 	        "\n");
 }
 
@@ -1774,6 +1776,8 @@ u32 MTUSize = 1450;
 #endif
 GF_ISOFile *file;
 Bool frag_real_time = GF_FALSE;
+Bool content_protection_in_adaptation_set = GF_FALSE;
+Bool content_protection_in_representation = GF_FALSE;
 Double mpd_update_time = GF_FALSE;
 Bool stream_rtp = GF_FALSE;
 Bool force_co64 = GF_FALSE;
@@ -3114,6 +3118,12 @@ Bool mp4box_parse_args(int argc, char **argv)
 		else if (!stricmp(arg, "-frag-rt")) {
 			frag_real_time = GF_TRUE;
 		}
+		else if (!stricmp(arg, "-cp-in-as")) {
+			content_protection_in_adaptation_set = GF_TRUE;
+		}
+		else if (!stricmp(arg, "-cp-in-rep")) {
+			content_protection_in_representation = GF_TRUE;
+		}
 		else if (!strnicmp(arg, "-dash-live", 10) || !strnicmp(arg, "-ddbg-live", 10)) {
 			dash_mode = !strnicmp(arg, "-ddbg-live", 10) ? GF_DASH_DYNAMIC_DEBUG : GF_DASH_DYNAMIC;
 			dash_live = 1;
@@ -3808,6 +3818,8 @@ int mp4boxMain(int argc, char **argv)
 		if (!e) e = gf_dasher_configure_isobmf_default(dasher, no_fragments_defaults, pssh_in_moof, samplegroups_in_traf, single_traf_per_moof);
 		if (!e) e = gf_dasher_enable_utc_ref(dasher, insert_utc);
 		if (!e) e = gf_dasher_enable_real_time(dasher, frag_real_time);
+		if (!e) e = gf_dasher_set_content_protection_in_adaptation_set(dasher, content_protection_in_adaptation_set);
+		if (!e) e = gf_dasher_set_content_protection_in_representation(dasher, content_protection_in_representation);
 		if (!e) e = gf_dasher_set_profile_extension(dasher, dash_profile_extension);
 
 		for (i=0; i < nb_dash_inputs; i++) {
--- a/include/gpac/media_tools.h
+++ b/include/gpac/media_tools.h
@@ -508,6 +508,18 @@ GF_Err gf_dasher_enable_utc_ref(GF_DASHSegmenter *dasher, Bool insert_utc);
  *	\return error code if any
 */
 GF_Err gf_dasher_enable_real_time(GF_DASHSegmenter *dasher, Bool real_time);
+/*Sets whether the ContentProtection element must be added to the AdapatationSet element or not.
+*	\param dasher the DASH segmenter object
+*	\param value if true, the ContentProtection element will be added to the AdaptationSet element.
+*	\return error code if any
+*/
+GF_Err gf_dasher_set_content_protection_in_adaptation_set(GF_DASHSegmenter *dasher, Bool value);
+/*Sets whether the ContentProtection element must be added to the Representation element or not.
+*	\param dasher the DASH segmenter object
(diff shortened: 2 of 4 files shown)


rbouqueau:
Hello,

Thanks for your contribution.

I added the ContentProtection element at the representation level in 9889e782163da80337d5fa124f3ba36364a2c9cd . Otherwise we lose compatibility with some devices (including the Chromecast).

I read the DASH IF IOP 3.1 (http://dashif.org/wp-content/uploads/2015/10/DASH-IF-IOP-v3.1.pdf). It doesn't forbid to repeat the information at the Representation level IMO. It shall be present at the AdaptationSet level and that's what we do. What do you think?

vladimir-kazakov:
Hi, Romain!

I think that proposed change is still valid.

You're right, 3.1 doesn't explicitly prohibit using ContentProtection in Representation. Actually, all previous versions don't do it as well. Only 2.0 says:

> In the case where license acquisition information is present in ContentProtection Descriptors, they shall only be present on the AdaptationSet element(s).

I guess this is what confused me initially. But now looking at it I don't think that this case is related to our case, because MP4Box doesn't include license acquisition information in ContentProtection.

The reason why I still think that the proposed change is valid is because all IOP versions explicitly tell about ContentProtection in AdaptationSet, but there's completely no information about having this element in Representation. Moreover, there's even information about why ContentProtection shall be in AdaptationSet. This is something that IOP expects clients to support. If they don't support this case, then they're not compatible with IOP. So, adding exceptions for some clients seems wrong to me.

Perhaps, the better solution would be a separate "advanced option" in MP4Box, that would control this behavior? Something that would do "include ContentProtection everywhere". In this case we could keep MPDs as small as possible and clean from exceptions, and at the same time, when it's really needed, we could add exceptions for specific clients.

rbouqueau:
I agree this can be confusing. Would you feel ok to report to DASH-IF so that they clarify?

vladimir-kazakov:
There's nothing to report to DASH-IF. What I was trying to say is that current logic of MP4Box seems wrong.

IOP requires clients to understand ContentProtection (CP) in AdaptationSet (AS). We also found that IOP doesn't prohibit CP in Representation (REP). In other words, a client that supports IOP **MUST** understand CP in AS and **MAY** use CP in REP.

You mentioned that CP in REP fixes playback for Chromecast. This means that Chromecast doesn't support IOP, because it doesn't understand CP in AS. Adding CP in REP fixes the playback, but Chromecast still doesn't support IOP. The problem is still there, but now it's hidden, and people may start thinking that Chromecast supports IOP, which is not true.

This is not a big issue if we think only about the playback. If the playback works, then who cares with which standard this media is compliant. But this is a big issue when we need media that we can use to test clients whether they support the specific standard. Since IOP requires CP only in AS, I think that MP4Box must follow this logic and not include anything else. This will allow content creators to produce clean (without any workarounds) media that can be used to test clients. The playback won't work on clients that don't support the specific standard (or interoperability point, or guidelines, or whatever else) making the problem visible and obvious. At the same time, if content creators don't care about such problems and just want the playback working everywhere, they could simply use "advanced options" to include workarounds for special devices.

Let's say, there would be 2 new advanced options: `-cp-in-as` and `-cp-in-rep`. When we would use an IOP profile without these advanced options, the MPD would contain only what is required by IOP (CP in AS). When we need to support Chromecast, we could simpy add `-cp-in-rep`, which would add CP in REP.

This would give us enough flexibility for 2 important cases:

1. Testing clients whether they support IOP.
2. Having playback on as much clients as possible.

Currently, only (2) is possible. And doing (1) is now problematic, because it requires manual editing of the MPD.

I could try to implement this proposal, but I need a green light :)

rbouqueau:
Thanks for your contribution and efforts :)

Maybe to precise why we are so conservative: there are many versions of the DASH-IF AVC/264, IOP, or whatever names and versions they represent. We clearly need a refresh (as you suggested in #41) to support them more appropriately when they are stable.

Maybe for Chromecast we should create a custom profile to get compatibility. And remove CP from Representation for the DASH-If profiles. Would it fit your need?

> Let's say, there would be 2 new advanced options

At least they exist programmatically. See ```content_protection_in_adaptation_set``` in 9889e782163da80337d5fa124f3ba36364a2c9cd.

Opinion?

vladimir-kazakov:
> Maybe for Chromecast we should create a custom profile to get compatibility.

No-no-no. Please, don't create more profiles. We already have more of them than needed :). There's no need in additional profiles. MP4Box just needs more flexibility in specifying where exactly the ContentProtection element must be located. See my last commit. I implemented what exactly I wanted.

The code should be self-explanatory, but here's an overview of the logic that I was trying to implement.

When the user specifies the profile or the interoperability point, the ContentProtection element will be located only in places that are explicitly required. If such place is undefined, then in the Representation.

But now the user may be more explicit about where the ContentProtection must be located. I added 2 new advanced options: `-cp-in-as` and `-cp-in-rep`. The user can use either one option or both options at the same time, depending on the user's needs.

Now, let's say the user needs to use the DASH-IF IOP and `-cp-in-rep`. In this case, the user violates the requirements of DASH-IF IOP. But if the user really wants to use DASH-IF IOP, he or she may add `-cp-in-as` as well.

Having these new options also allows the user to do something that wasn't possible before. For example, the user may want to use the DASH profile, but put ContentProtection element in AdaptationSet. So, he or she just need to specify: `-profile onDemand -cp-in-as`.

jeanlf:
Thanks for the patch. I still think having a profile for chromecast makes sense, just like we have profiles for DASH-IF IOPs: we don't expect users to understand the impact of all possible options for DASH generation, profiles make life much simpler for them.

jean

vladimir-kazakov:
> I still think having a profile for chromecast makes sense, just like we have profiles for DASH-IF IOPs

DASH-IF IOP is written down in a well-known, publicly available document. It's like a standard. Speaking of custom profiles, nobody knows about them and nobody knows what to expect from them. How are you going to support them? Are you going to create custom profiles for every Google's device that doesn't support standards?

I guess it's a wrong path, which will also overcomplicate MP4Box.

> we don't expect users to understand the impact of all possible options for DASH generation

Then how they will be choosing a profile from the long list of custom profiles, requirements for which are not even defined anywhere? :) Moreover, a custom profile would limit people. For example, you've got support for Chromecast, but what if there's yet another device that needs a special treatment? Custom profiles won't solve this problem. But if MP4Box will provide flexibility to shape MPDs however people want, then people will be able to add support for special devices without waiting for changes in MP4Box.

jeanlf:
sorry for the delay reply - I understand your point of view but if we really want to go that way we would just get rid of all DASH-IF profiles since they all are "custom profile", at least according to DASH which is what MP4Box support :)

Anyway that's all a question of taste, what matters is the ability to select where to put the CP element. I won't got for 2 switches but rather a single one (in AS, in REP or in BOTH), I'll merge and update

vladimir-kazakov:
> if we really want to go that way we would just get rid of all DASH-IF profiles since they all are "custom profile"

Well, I wouldn't say that DASH-IF IOP is a custom profile. At least not in a way I see it. I mean, there are some profiles, like DASH-IF IOP, HbbTV, DVB-DASH, that are well-defined. So, we know what to expect from them, which means that they can be supported.

At the same time, proposed mixture of DASH-IF IOP and hacks for Chromecast is not defined anywhere. That's what truly looks like a "custom profile" to me. If someone like DASH-IF will be able to define such profile and will publicly release the guidelines / spec / or whatver it will be called, then sure, why not to support it, as long as it's popular and meaningful :).

> I won't got for 2 switches but rather a single one (in AS, in REP or in BOTH), I'll merge and update

I was also thinking about single parameter, but for some reasons chose two parameters. At the same time, by looking at your changes, I actually like the single parameter better. Although I'd also note that there may be the third location of CP - Sub-Representation, which makes the BOTH option a bit wrong in a long term.

Also, as I mentioned in my comments for the commit, it would be really nice if choosing a profile would set some correct defaults. For example, if we know that DASH-IF IOP requires CP in AS, then there's no point to set CP mode to AS, unless we want to override something.

PR #377, opened 2016-01-18
GLES2 - disable mode

emmanouil:


modules/droid_out/droid_vout.c
```diff
@@ -359,90 +605,6 @@ int createTexture(AndroidContext *rc)
...
-
-//	glGetRenderbufferParameterivOES(GL_RENDERBUFFER_OES, GL_RENDERBUFFER_WIDTH_OES, &backingWidth);
-//	glGetRenderbufferParameterivOES(GL_RENDERBUFFER_OES, GL_RENDERBUFFER_HEIGHT_OES, &backingHeight);
-
-//	LOG( ANDROID_LOG_ERROR, TAG, "Android Depth Buffer Size: %dx%d\n", backingWidth, backingHeight));
-
-//    glRenderbufferStorageOES(GL_RENDERBUFFER_OES, GL_DEPTH_COMPONENT16_OES, rc->width, rc->height);
-
-//    glFramebufferRenderbufferOES(GL_FRAMEBUFFER_OES, GL_DEPTH_ATTACHMENT_OES,
-//            GL_RENDERBUFFER_OES, rc->depthbuff);
```
rbouqueau: Is this dead code or do we plan to enable it somewhen?

emmanouil: I definitely do not - but to be on the safe side, I only removed what was between "GLES_FRAMEBUFFER_TEST" defs

PR #424, opened 2016-02-18
Android video driver cleanup

canatella:
Cleanups for the Android video driver.

- Logging is now using GPAC logging api, it makes it easier to collect all the logs in the same place.
- OpenGL error checking is using an inline function instead of a macro. It enforces semicolon use and shows that the gl_check_error() statement is in fact doing something.
- Fixed some indentation problems.


jeanlf:
merged in 441, thanks

PR #425, opened 2016-02-20
Fixed UTF8 memory size calculation

kazuki:
In 108b820(master, 2016/2/19) nightly build binary.

```
> mp4box.exe -info 日本語漢字.mp4
Length allocated for conversion of wide char to UTF-8 not sufficient
```

many cjk charactors encoded to three bytes in utf8 encoding.
But current implementation, allocate 2 * len bytes.

[RFC3629: UTF-8, a transformation format of ISO 10646](https://tools.ietf.org/html/rfc3629) defines UTF8 max octet sequence size is 4.

This patch fixed utf8 memory size calculation.


cconcolato:
Sorry for this oversight and thank you for the fix. It is now merged.

Issue #432, opened 2016-02-23
Building 0.6.0 fails when 0.5.0 is already installed


ryandesign:
Building gpac 0.6.0 fails when gpac 0.5.0 is already installed. Here is the error I see on OS X when building gpac with MacPorts:

```
/usr/bin/clang -L/opt/local/lib -framework Carbon -o ../../bin/gcc/MP4Client main.o extract.o carbon_events.o -L../../bin/gcc  -lgpac
Undefined symbols for architecture x86_64:
  "_AVI_set_audio", referenced from:
      _avi_audio_frame in extract.o
  "_AVI_write_audio", referenced from:
      _avi_audio_frame in extract.o
  "_gf_cfg_discard_changes", referenced from:
      _mp4client_main in main.o
  "_gf_fclose", referenced from:
      _GPAC_EventProc in main.o
      _mp4client_main in main.o
      _init_rti_logs in main.o
      _write_bmp in extract.o
      _write_png in extract.o
      _write_depthfile in extract.o
      _write_texture_file in extract.o
      ...
  "_gf_fopen", referenced from:
      _mp4client_main in main.o
      _init_rti_logs in main.o
      _write_bmp in extract.o
      _write_png in extract.o
      _write_depthfile in extract.o
      _write_texture_file in extract.o
      _write_raw in extract.o
      ...
  "_gf_free", referenced from:
      _mp4client_main in main.o
      _write_png in extract.o
      _dump_file in extract.o
  "_gf_lang_find", referenced from:
      _ViewOD in main.o
  "_gf_lang_get_name", referenced from:
      _ViewOD in main.o
  "_gf_malloc", referenced from:
      _mp4client_main in main.o
      _write_png in extract.o
      _dump_file in extract.o
  "_gf_net_get_utc", referenced from:
      _mp4client_main in main.o
      _on_gpac_log in main.o
  "_gf_net_set_ntp_shift", referenced from:
      _mp4client_main in main.o
  "_gf_sc_add_audio_listener", referenced from:
      _dump_file in extract.o
  "_gf_sc_remove_audio_listener", referenced from:
      _dump_file in extract.o
  "_gf_sha1_csum", referenced from:
      _write_hash in extract.o
      _dump_depth in extract.o
      _dump_frame in extract.o
  "_gf_sys_clock_high_res", referenced from:
      _on_gpac_log in main.o
      _init_rti_logs in main.o
  "_gf_sys_set_args", referenced from:
      _mp4client_main in main.o
[...]

PR #433, opened 2016-02-23
Specify $(LDFLAGS) after, not before, local -L flags

ryandesign:
Move `$(LDFLAGS)` to the end. Fixes gpac/gpac#432.

This PR was created by running this command:

```
find . -name Makefile -print0 | xargs -0 sed -i '' -E 's%( \$\(LDFLAGS\))(.*)%\2\1%g'
```

Note this is BSD sed syntax, not GNU sed.

This may not be exactly what you want. Although this fixes the build failure I experienced, it doesn't address the fact that `$(EXTRALIBS)` still comes before local library directory flags, and `$(EXTRALIBS)` might contain either local library directory flags or system library directory flags.

### DIFF
--- a/applications/dashcast/Makefile
+++ b/applications/dashcast/Makefile
@@ -65,7 +65,7 @@ SRCS := $(OBJS:.o=.c)
 all: $(PROG)
 
 DashCast$(EXE): $(OBJS)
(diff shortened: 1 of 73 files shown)


jeanlf:
merged and working thanks a lot

Issue #230, opened 2015-09-07
Warning: gpac is too old - Using --static-mp4box then OpenGL support: no


ghost:
When trying to compile x264 against gpac, unless you build gpac as `--static-mp4box` then x264 complains that the gpac version is too old, but by doing this, then the OpenGL support is no...

How do we get OpenGL support in gpac and compile it also against x264 without it complaining?

rbouqueau:
Hi,

Can you confirm you are talking about GPAC OpenGL support? GPAC is made of several tools, including a packaging and a playback libraries. x264 can use the packager, which has almost no dependency. Only the playback library takes advantage of OpenGL. So you don't need OpenGL to link against x264.

Also when do you get the "Warning: gpac is too old" warning? There was a similar issue a year ago, we could find a way to reproduce with the latest versions of x264 and gpac. I'd like us to fix it.

Geyup:
I really don't know what OpenGL support this is, in GPAC or the System, all I know is the ```** Detected libraries **``` when you compile GPAC as ```--static-mp4box``` then it shows ```OpenGL support: no``` like below;


** Detected libraries **
zlib: system
OSS Audio: yes
ALSA Audio: yes
Jack Audio: no
PulseAudio Audio: no
DirectFB support: no
X11 Shared Memory support: yes (path: /usr/X11R6)
X11 XVideo support: yes
SDL Support: yes
```OpenGL support: no```
TinyGL support: no
OpenSSL support: no
Mozilla XUL/GECKO support: no
DVB Support: yes
XMLRPC Support: no
wxWidgets support: no

I just grabbed the latest source from git;

```git clone https://github.com/gpac/gpac.git```

These are my compile options;
```
LDFLAGS="-L/usr/lib${LIBDIRSUFFIX}" \
./configure \
  --prefix=/usr \
  --mandir=/usr/man \
  --disable-opt \
  --cpu=${ARCH} \
  --static-mp4box \
  --extra-cflags="$SLKCFLAGS" \
  $wx_option
```

I also grabbed the latest version of x264;
```ftp://ftp.videolan.org/pub/x264/snapshots/last_x264.tar.bz2```

If I compile gpac with ``` --static-mp4box ``` there is no warning message, if I remove it, then x264 complains.

rbouqueau:
Ok so it seems to come from the x264 detection. Have you reported the issue to their developers?

Geyup:
No, wasn't thinking about them yet.

[...]

PR #489, opened 2016-04-13
Pkg config private libs

### LINKED ISSUE #5: check pkgconfig for static libgpac
Apparently we don't fill it correctly. Also see https://sourceforge.net/p/gpac/bugs/341/

rbouqueau:
Needed to fix #5. Then:
 - [ ] create a new tag v0.6.2
 - [ ] propose a patch to x264 for #230


rbouqueau:
Can we merge?

PR #495, opened 2016-04-19
Fix compilation with `CONFIG_DIRECTX=yes`.

XhmikosR:
I was getting build errors with this commandline:

```
bash ./configure --cpu=i686 --enable-static-bin --static-mp4box --extra-cflags='-m32' --extra-ldflags=-s
```

### DIFF
--- a/static.mak
+++ b/static.mak
@@ -181,13 +181,13 @@ CFLAGS+=-DGPAC_HAS_JP2
 endif
 
 ifneq ($(CONFIG_FT), no)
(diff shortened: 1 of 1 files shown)


XhmikosR:
BTW, this failed for me with

```
bash ./configure --cpu=i686 --enable-static-bin --static-mp4box --extra-cflags='-m32' --extra-ldflags=-s
```

jeanlf:
indeed thanks for the report

PR #552, opened 2016-06-14
terminal/scene: replace non-breaking spaces added in f5d1997

wiiaboo:
They broke at least MinGW builds.


wiiaboo:
This doesn't fix build completely: https://i.fsbn.eu/Cjox.txt

jeanlf:
this hould have already been fixed, thanks for the report.

PR #586, opened 2016-07-12
fix IsEncrypted in CENC xml: write only the PSSH

rbouqueau:
When merged, I need to modify http://gpac.io/mp4box/encryption/common-encryption/


jeanlf:
doc updated

PR #589, opened 2016-07-18
android osmo4: MediaCodec video decoder 

rauf:


modules/mediacodec_dec/mediacodec_dec.c
```diff
@@ -0,0 +1,891 @@
...
+{
+    return GF_NOT_SUPPORTED;
+}
+
+
+void convertNV12toNV21(char *bytes, u32 width, u32 height) 
+{
+    u32 i;
+    for(i = width*height; i < 3 * (width * height)/ 2 - 1; i++){
+        bytes[i] = bytes[i + 1];
```
rbouqueau: I think we should replace this with proper nv21 support.
@jeanlf ?

rbouqueau:
I made a quick review. Seems ok to me except the comment I made.

modules/mediacodec_dec/mediacodec_dec.c
```diff
@@ -0,0 +1,891 @@
...
+u8 sdkInt() 
+{
+    char sdk_str[3] = "0";
+    __system_property_get("ro.build.version.sdk", sdk_str, "0");
+    return atoi(sdk_str);
+}
+#endif
+
+
+u32 androidPixelFormat(u32 pix_fmt) {
```
RodolpheFouquet: I think we found the mapping on these values in the Android doc, maybe we should add a comment to precise that.

modules/mediacodec_dec/mediacodec_dec.c
```diff
@@ -0,0 +1,891 @@
...
+        LOGE("AMediaFormat_new failed");
+        return GF_CODEC_NOT_FOUND;
+    }
+
+
+#ifdef GPAC_ANDROID
+
+        ctx->pix_fmt = GF_PIXEL_NV21;
+
+switch (ctx->esd->decoderConfig->objectTypeIndication) {
```
RodolpheFouquet: Shouldn't we split these cases into functions? MCDec_InitHevcDecoder, ...InitAvcDecoder, ...

rauf: sure i will

rauf: @RodolpheFouquet  sure i will

RodolpheFouquet:
Except 2 remarks I have done, it sounds good to me

modules/mediacodec_dec/mediacodec_dec.c
RodolpheFouquet: I think it should just land to a minor modification of the current shader, right?

rbouqueau: There is no support for NV12 in GPAC yet. We could take every NV21 occurence and add NV12 ; there doesn't seem to be that much occurrences: https://github.com/gpac/gpac/search?utf8=%E2%9C%93&q=nv21
@rauf Is it clear for you?

rauf: @rbouqueau  While going through the formats, I tried YPVU since it is VU interleaved and U and V are swapped, and it works. The colors are accurate. There is no need to add NV12 as of now, we can change the format to GF_PIXEL_YPVU :)

rbouqueau: excellent :)

rauf:
removed setting of KEY_COLOR_FORMAT as it was only necessary for encoders. ref: https://developer.android.com/reference/android/media/MediaFormat.html

RodolpheFouquet:
My tests seem to be conclusive, except for 4k (though it wasn't running before). I need to test with more content to test mpeg4 part 2 and hevc streams.

@rbouqueau Can you check:
- that it runs on your phone
- that https://github.com/rauf/gpac/blob/a776671c9ec72997cb0b5b05bff02b1ac8ee1908/modules/mediacodec_dec/mediacodec_dec.c seems ok?

rbouqueau:
Thanks

@RodolpheFouquet @rauf 

1) The onfig file editor doesn't work here. It displays a generic page. It seems like GPAC complains about GPAC.cfg file rights, may it be related?

2) Playback: I see some weird colors at the top of the video capture by the smartphone camera. Do we have a set of videos for the validation?

3) There are some defines, but I don't think it compiles without them. Am I correct?
https://github.com/rauf/gpac/blob/a776671c9ec72997cb0b5b05bff02b1ac8ee1908/modules/mediacodec_dec/mediacodec_dec.c#L33

4) Hardcoded value: does it have an influence?
https://github.com/rauf/gpac/blob/a776671c9ec72997cb0b5b05bff02b1ac8ee1908/modules/mediacodec_dec/mediacodec_dec.c#L390

RodolpheFouquet:
Wow I missed the hardcoded value for the framerate, that could explain my issues with the 4k playback since it was a 15fps stream.

RodolpheFouquet:
3) There are some defines, but I don't think it compiles without them. Am I correct?
https://github.com/rauf/gpac/blob/a776671c9ec72997cb0b5b05bff02b1ac8ee1908/modules/mediacodec_dec/mediacodec_dec.c#L33

<== it would not I think, unless the makefile is Android specific

rbouqueau:
> <== it would not I think, unless the makefile is Android specific

So I propose to remove the defines. I remember we initially introduced them so that the module core structure could be tested outside Android. That's useless now.

rauf:
Hi,

@rbouqueau @RodolpheFouquet 

1)  Config File Editor should work. Rodolphe and I both have tested it. Can you give more info on this?

2) I have seen this issue on stackoverflow. Which stream and device was it?

3) Yes, they were introduced so that I can develop the module without Android on Linux. I will remove them in next commit.

4) Nice, I also think that would be the issue with Rodolphe's 4k 15fps video, since the decoder currently play 4k video from my device fine, will remove it in next commit.

Thanks.

rbouqueau:
1) 2) It must come from my phone: Motorola G 2015 (Android 6.0.1). I see the message "Error\n Core System failure in opening GPAC.cfg in mode wt: 30". I sometimes also see a shader error - weird.

3) 4) Thanks.

rauf:
@rbouqueau  

Try uninstalling and reinstalling the app. This will solve shader issue. 
Also, just ignore the error and try to use config file editor.

rbouqueau:
@rauf Thank you. The shaders are installed by I think I get the error because the gpac terminal is recreated before the old one is deleted. I'll have a look when I have time.

For me the issue of non writable gpac.cfg files is a serious problem...

PR #606, opened 2016-08-07
Fix broken compilations with MSYS MinGW since 5b14f8328

wiiaboo:
The referenced commit broke compilations when both /mingw32 and
/mingw64 paths exist and people tried to compile 32-bit.


rbouqueau:
Thank you

Issue #631, opened 2016-09-09
m3u8 to mpd conversion not working when EXT-X-KEY option is present


AdithyanIlangovan:
Thanks for reporting your issue. Please make sure these boxes are checked before submitting your issue - thank you!

- [x] I looked for a similar issue and couldn't find any.
- [x] I tried with the latest version of GPAC. Installers available at http://gpac.io/downloads/gpac-nightly-builds/
- [x] I give enough information for contributors to reproduce my issue (meaningful title, github labels, platform and compiler, command-line ...).
Detailed guidelines: http://gpac.io/2013/07/16/how-to-file-a-bug-properly/

I am trying to convert a .m3u8 manifest to .mpd manifest. 
For this purpose, I use the following command :
> MP4Box -mpd master.m3u8

You can find the master.m3u8 file attached here [[master.m3u8.zip](https://github.com/gpac/gpac/files/463897/master.m3u8.zip)].

Normally, this works. But whenever the .m3u8 file has a **EXT-X-KEY** method in it, the command does not work. That is whenever such an option is present:
>#EXT-X-KEY:METHOD=AES-128,URI="https://priv.example.com/key.php?r=52",  \
IV=0x9c7db8778570d05c3177c349fd9236aa

The code crashes, by generating an half-completed master.mpd file : [master.mpd.zip](https://github.com/gpac/gpac/files/463909/master.mpd.zip)

The log file is attached below : 
[mp4box-bug-log.txt](https://github.com/gpac/gpac/files/463913/mp4box-bug-log.txt)

The backtrace from gdb: 
>(gdb) backtrace
>#0  0x00000000005b5af0 in gf_mpd_print_segment_list ()
>#1  0x00000000005bb3fa in gf_mpd_write_file ()
>#2  0x0000000000413782 in mp4boxMain ()
>#3  0x00007ffff75247ed in __libc_start_main (main=0x40b080<main>,argc=3,ubp_av=0x7fffffffdff8,init=<optimized out>,fini=<optimized out>,rtld_fini=<optimized out>,
stack_end=0x7fffffffdfe8) at libc-start.c:226                                                                                                                           
>#4  0x000000000040b0b1 in _start () 

I am happy to provide other supporting information.

Edit 1 : I took the sample m3u8 from apple official site : https://developer.apple.com/library/ios/technotes/tn2288/_index.html.

jeanlf:
Fixed on git thanks to your comments !

PR #632, opened 2016-09-09
Bug fig - Infinite loop condition.

AdithyanIlangovan:
- The variable "i" was not incremented in the for loop. I have fixed this.
- Also I am not sure, in the line 2042:

> fprintf(out, "%02x", url->key_iv[i]);

whether that should be "i" or "idx". I request the maintainer to check this.


AdithyanIlangovan:
This is the bug fix for issue number #631.
I had reported this issue.

jeanlf:
Thanks for the report, fixed on git

PR #638, opened 2016-09-21
MMT MPU Generation

DerouineauNicolas:
Tested with 

MP4Box -add ~/Vidéos/small.mp4 -mmt-seq-num 10 -mmt-asset-id-scheme URI -mmt-asset-id-value http://www.gpac.io mpu10.mp4

MP4Box -add ~/Vidéos/small.mp4 -mmt-seq-num 12 -mmt-asset-id-scheme URI -mmt-asset-id-value http://www.gpac.io mpu10.mp4

If the input contains N tracks, 

MP4Box -add ~/Vidéos/VideoandAudio.mp4 -mmt-autogen -frag 100 mpu.mp4

N Files will be generated. For this case, UUID are Hardcoded values . 

Once generated, mpu boxes can be parsed with 

MP4Box -std -diso mpu11.mp4


applications/mp4box/main.c
```diff
@@ -1850,6 +1852,12 @@ const char *grab_m2ts = NULL;
 const char *grab_ifce = NULL;
 #endif
 FILE *logfile = NULL;
+Bool mpu_is_complete=0;
```
cconcolato: Why use global variable ?

DerouineauNicolas: Is there another way to retrieve the parameters found in mp4box_parse_args function ?

jeanlf: nope that's the right way...

rbouqueau:
1) looks like this variable is unused, is this expected?
2) can't we put this global vars as static to limit the scope?

Romain Bouqueau
CEO @ www.gpac-licensing.com
romain.bouqueau@gpac-licensing.com
+33 (0) 614 685 991
skype : romain.bouqueau

2016-09-21 23:34 GMT+02:00 jeanlf <notifications@github.com>:

DerouineauNicolas:
@rbouqueau: 
1, Yes, this parameter is not used at the moment, as this is only preliminary investigation.
2, You're right, I'll set them as static

PR #641, opened 2016-09-22
Introduced 360video sensor-based navigation for Android

emmanouil:
Introduced sensor-based navigation for 360videos on Android. The navigation is based on Rotation Vector composite sensor measurements, and if it's unavailable, rolls back to Accelerometer and Magnetic Field.

INCLUDES:
- Customizable Low-pass filter for smoothing (on Mag+Acc Orientation values)
- Customizable measurement threshold (pseudo-high-pass filter on Mag+Acc Orientation values)

TODO:
 - Lock to landscape orientation, when in 360
 - Stereoscopic support

Known Issues:
 - Sensor data gathering does not pause when in background
 - Unpredictable behaviour when the navigation stars in portrait mode (however, switching to it afterwards is ok)


SrujithPoondla:
are these features working?? Please add documentation for using these features, I couldn't find any one of these features in the present build.

emmanouil:
these features do work.
To test, you should have a compatible file and when you open it you should:
either tap and hold until the menu pops-up and then select "View as 360"
 or
add #LIVE360TV (if you are using the URI file open)
NOTE: this sensor-based navigation will only work in Landscape mode and works better if your device has a gyroscope

SrujithPoondla:
Thanks for the info @emmanouil. I am able to find the menu but I couldn't find the view as 360 option anywhere. Could you please add some steps so that it will be helpful.

jeanlf:
you need to use the built-in GUI (located at the bottom of the screen) and not the native android menu. Click open file icon, locate the file you want and long click (> 2s) on it, then select "View 360"

note: use a recent build from the previous week (rev 1090 or more), older ones seem to have an issue loading the GUI

SrujithPoondla:
Thanks @jeanlf. I followed above steps, I entered the url and the file got added to the playlist. But i am unable to find the 360 view option even though i long clicked the file in the playlist. Is there any other way i can switch to 360 view.

jeanlf:
doesn't work in the playlist manager, only in the file open rom the main gui

SrujithPoondla:
@jeanlf thanks for the info. It is working :)

SrujithPoondla:
Does Gpac uses openGL to render all the tiles to a 360 degree video frame? How it is transforming the video to 360 degree if we add #LIVE360TV at the end of URL ? Could some one help in this issue

jeanlf:
please don't cross post  - same as #710

PR #664, opened 2016-10-14
Get the views info from the vps to fix stride handling for MVHEVC vs SHVC

jeljeli:


jeanlf:
supported in master (and just cleaned up), closing

PR #696, opened 2016-11-16
Fix output of unsigned offset, size fields

porcelijn:
I ran into negative chunk entry offsets in `stco` box when testing `MP4Box -diso` with large mp4 files (>4GB).
I took the opportunity to also fix signedness for a couple of size, duration and index fields.


jeanlf:
thanks !

Issue #711, opened 2016-12-06
Android mediadec_dec module crashes on some streams


canatella:
On my device, all HLS and some Dash streams crash Osmo4.

Device: OnePlus 2, Android 6.0.1
Sample streams: 
- http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8
- http://rdmedia.bbc.co.uk/dash/ondemand/bbb/2/client_manifest-common_init.mpd

GPAC Nightly: osmo4-0.6.2-DEV-rev1172-g3782ef6-master.apk

Log:

12-06 17:13:51.313  8786  8786 D Osmo4   : Copy assets of shaders
12-06 17:13:51.533  8786  8803 I GPAC_WRAPPER: Selected JNI VERSION[0]
12-06 17:13:51.533  8786  8803 I GPAC_WRAPPER: Registering com/gpac/Osmo4/GPACInstance natives
12-06 17:13:51.533  8786  8803 I GPAC_WRAPPER: Registering 12 methods...
12-06 17:13:51.533  8786  8803 I GPAC_WRAPPER: Registering natives DONE, now registering pthread_keys with destructor=0xdd648309
12-06 17:13:51.534  8786  8803 I GPAC_WRAPPER: Initializing GPAC with URL=http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8...
12-06 17:13:51.534  8786  8803 I GPAC_WRAPPER: GPAC.cfg found in /storage/emulated/0/osmo/, force using it.
12-06 17:13:51.534  8786  8803 I GPAC_WRAPPER: Loading User Config GPAC.cfg...
12-06 17:13:51.535  8786  8803 I GPAC_WRAPPER: loading modules in directory /data/data/com.gpac.Osmo4/lib...
12-06 17:13:51.537  8786  8803 D GPAC_WRAPPER: Loading GPAC terminal, m_user=0xdf753014...
12-06 17:13:51.538  8786  8803 V GPAC_WRAPPER: SetupLogs()
12-06 17:13:51.564  8786  8807 I GPAC_WRAPPER: Rebuilding methods for thread -582481616
12-06 17:13:51.564  8786  8807 I GPAC_WRAPPER: Getting current Thread -582481616...
12-06 17:13:51.564  8786  8807 I GPAC_WRAPPER: Getting current Thread DONE = 0xdd599118, now registering before exit...
[...]

PR #717, opened 2016-12-12
Improved mediacodec performances

jeljeli:
Decoding avc streams ( mp4, ts, dash, hls)

TODO :
- Fix Cropping
- HEVC decoding


jeljeli:
cf #711

PR #722, opened 2016-12-20
[Android] No need to fill audio buffer in a critical region

jeljeli:
This commit fixes crash when we enable debug mode


canatella:
Could you extend a bit on the reason for this ? Get*ArrayElements will copy the array instead of a direct access using GetPrimitiveArrayCritical. 

Is the performance gain minimal when using GetPrimitiveArrayCritical ? And what crashes does it fixes ? I have code here which dispatch an event when entering and exiting the critical section so that other threads can be aware of it and acts accordingly. Maybe the performace gain of direct access vs copy does not justify this but I'd like to have some insight.

canatella:
Indeed, thank you for the information!

PR #739, opened 2017-01-18
Change TTML namespace tests

fcartegnie:
TTML Namespace test was not testing namespace.

Untested: Unsure if those files need to be referenced somewhere else.

### DIFF
--- a/tests/media/ttml/ebu-ttd_sample_mixed_namespaces.ttml
+++ b/tests/media/ttml/ebu-ttd_sample_mixed_namespaces.ttml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<tt xmlns:ttp="http://www.w3.org/ns/ttml#parameter"
+    xmlns:tts="http://www.w3.org/ns/ttml#styling"
+    ttp:timeBase="media"
+    ttp:cellResolution="50 30"
+    xml:lang="de" xmlns="http://www.w3.org/ns/ttml" xmlns:tt="http://www.w3.org/ns/invalidttml">
+	  <head>
+		    <styling>
+			      <style xml:id="defaultStyle" tts:fontFamily="Verdana, Arial, Tiresias"
+                tts:fontSize="160%"
+                tts:lineHeight="125%"
+                tts:color="#ffffff"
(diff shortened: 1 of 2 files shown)


rbouqueau:
IIUC the mixed namespace is correct and both ```<p>``` elements should be imported. Are we on the same line?

Explanation from https://www.w3.org/TR/ttml2/#vocabulary-namespaces :

> If a reference to an element type is used in this specification and the name of the element type is not namespace qualified, then the TT Namespace applies.

fcartegnie:
no. they should all fail.
vocabulary " indicates the default prefix used "
In the mixed example, the default namespace is bound to TTML, and the tt prefix is bound to another invalid namespace, thus not belonging to TTML.

tests/media/ttml/ebu-ttd_sample_mixed_namespaces.ttml
```diff
@@ -0,0 +1,28 @@
...
+                tts:textAlign="center"/>
+		    </styling>
+		    <layout>
+			      <region xml:id="bottom" tts:origin="10% 10%" tts:extent="80% 80%"
+                 tts:displayAlign="after"/>
+		    </layout>
+	  </head>
+	  <body>
+		    <div style="defaultStyle">
+			      <tt:p begin="00:00:00.000" end="00:00:01.400" region="bottom" xml:id="sub1">00:00:00.000<br/>00:00:01.400</tt:p>
```
nigelmegitt: These two `tt:p` elements are not in the TTML namespace; since this is an EBU-TT-D example file it is invalid for two reasons:

1. The EBU-TT-D spec does not allow content from other namespaces except in metadata
2. The cardinality of the `p` element as a child of `div` is `1..*`; this `div` has zero child TTML `p` elements (i.e. `p` elements in the correct namespace) so it is not conformant. The `p` elements in the non-TTML namespace do not count here.

nigelmegitt:
@fcartegnie wrote:
> no. they should all fail.

I agree!

rbouqueau:
I agree too, thanks for taking the time for the explanation. However before accepting the PR, we need to make the appropriate fixes ; otherwise our buildbot will send us a lot of emails ;)

PR #753, opened 2017-02-08
Fix relative paths

aureliendavid:
This PR does some dusting in the handling of relative paths. 

Most notably, it fixes the following scenario: 

the test file `tests/media/bifs/bifs-2D-background-background2D-movie.bt` contains 

```
      muxInfo MuxInfo {
       fileName "../auxiliary_files/enst_video.h264"
      }
```

if we export it to xmt with 

`mp4box -xmt media/bifs/bifs-2D-background-background2D-movie.bt`

it will wrongly try to "resolve" the relative path, resulting in 

`        <StreamSource url="media/auxiliary_files/enst_video.h264" >`

After that, doing

` MP4Box -mp4 media/bifs/bifs-2D-background-background2D-movie.xmt`

will fail because it's trying to open `media/bifs/media/auxiliary_files/enst_video.h264` instead of  `media/auxiliary_files/enst_video.h264`

With this PR, the same command outputs: 

`        <StreamSource url="../auxiliary_files/enst_video.h264" >`

and the next command can run.

### DIFF
--- a/src/odf/odf_dump.c
+++ b/src/odf/odf_dump.c
@@ -1697,11 +1697,7 @@ GF_Err gf_odf_dump_muxinfo(GF_MuxInfo *mi, FILE *trace, u32 indent, Bool XMTDump
 {
 	char *full_path = NULL;
 
-	if (mi->file_name && mi->src_url) {
-		if (strchr(mi->src_url, '/') || strchr(mi->src_url, '\\')) {
-			full_path = gf_url_concatenate(mi->src_url, mi->file_name);
-		}
-	}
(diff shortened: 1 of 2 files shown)


rbouqueau:
Hey Aurelien! Thanks.

FYI for the formatting part, I suggested a while ago to rely on ```AStyle```, see https://github.com/gpac/gpac/blob/master/doc/CODING_STYLE . I suggested to @jeanlf to make a pass at least at each release!

aureliendavid:
thanks @rbouqueau I'll make a pass of it on another commit and try to stick to this style from now on

PR #759, opened 2017-02-22
small things for mobile builds

aureliendavid:
two things I came across while rebuilding the project on new machines: 

android: a return type warning seem to give an error (Werror is probably defined somewhere)

ios: when ran for the first time, the xcodebuild archive command complains that no schemes exists in the project

schemes can be auto-generated in the xcode gui, but for the project to be buildable from scratch from the command line, it's better to include the schemes in the repo


sandyradhia:
@aureliendavid hello . please may u explain me how to integrated mp4box on android projet to build an audio live stream ? thanks i'm fighting since 8 month today with gpac i already build libgpac.so and libmp4box.so but when call native void run in my projet it don't work error occur native methode not found . thanks

PR #784, opened 2017-03-24
Merge mpd struct generation branch in master

DerouineauNicolas:
This PR is storing the mpd elements in a structure contained in the dasher and prints the values at the end of segmentation.

The integration is tested with:
dash.sh: OK
dash-live.sh: OK
dash-live-generation.sh: OK
dash-if-live.sh: time out on test, to be fixed
dash-if-ondemand.sh: Ok, but 2 files pointers are still open in playback, to be fixed.

### DIFF
--- a/src/utils/os_divers.c
+++ b/src/utils/os_divers.c
@@ -1722,6 +1722,24 @@ s32 gf_net_get_ntp_diff_ms(u64 ntp)
 	return (s32) (local - remote);
 }
 
+GF_EXPORT
+u64 gf_net_get_ntp_ms()
+{
+	u32 sec, frac;
+	u64 time_ms;
+	Double msec;
+
+	gf_net_get_ntp(&sec, &frac);
+	time_ms = sec;
+	time_ms *= 1000;
+	msec = frac*1000.0;
+	msec /= 0xFFFFFFFF;
+
+	time_ms += (u32)msec;
+
+	return time_ms;
+}
+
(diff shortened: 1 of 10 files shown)


src/utils/os_divers.c
```diff
@@ -1722,6 +1722,24 @@ s32 gf_net_get_ntp_diff_ms(u64 ntp)
...
+	time_ms = sec;
+	time_ms *= 1000;
+	msec = frac*1000.0;
+	msec /= 0xFFFFFFFF;
+
+	time_ms += (u32)msec;
+
+	return time_ms;
+}
+
```
rbouqueau: Duplicate of gf_net_get_utc()

DerouineauNicolas: I might be wrong, but it seems there is a difference between gf_net_get_utc and gf_net_get_ntp_ms.

gf_net_get_utc is subtracting GF_NTP_SEC_1900_TO_1970 from  gf_net_get_ntp, whereas gf_net_get_ntp_ms does not.

rbouqueau: You're right. I'm just trying to remove duplicate code.

ronak2121:
Hi,

I"m wondering what's going on with this issue; as I'm impatiently waiting for support for HLS Manifest generation from Fragmented MP4s. Was curious when this would be merged and finished.

Thanks

Issue #616, opened 2016-08-31
gpac fails to build with openssl-1.1


Polynomial-C:
Latest git head fails like this:

```
[...]

PR #829, opened 2017-05-18
Fix build with OpenSSL 1.1

yan12125:
#616 is partially fixed with this PR. Deprecated functions in OpenSSL are not handled yet.

By the way, wget uses a similar fix: http://git.savannah.gnu.org/cgit/wget.git/commit/src/openssl.c?id=309e72c74f4ea6c3707ca185bc92ae95ebf56b16

Test environment: Arch Linux x86_64


aureliendavid:
included in e64a7d2

PR #844, opened 2017-06-07
[Android] HEVC support + multiple MediaCodec instances 

jeljeli:
- Added hevc support in mediacodec_dec
- Added the ability to instantiate multiple hardware decoders simultaneously ( tested with tiled 3x3 4K H264 stream in HTC 10)


jeljeli:
if an android device have the "adaptive-playback" capability we don't need to parse XPSs and reconfigure decoder when a change is detected. Let the decoder do it.

PR #861, opened 2017-06-28
Remove WebVTT terminators before boxstring creation

porcelijn:
This pull request removes the final trailing whitespaces (spec. newline and carriage return, WebVTT terminators) from input before wrapping it as a text/wvtt boxstring type.
Various drafts on BMFF WebVTT (ISO/IEC-14496-30) seem to mandate (or at least imply) that the **boxstring** text (payl,vttC,iden,vtta,sttg,...) must not include any trailing newline/CRs present in the WebVTT (plain text) syntax. (See: w13040, w13483)

I noticed though, that a roundtrip like:
```
$ MP4Box -add subs.vtt -new part30.mp4
$ MP4Box -raw part30.mp4
```
unfortunately does duplicate most of the newlines. While this is mostly a cosmetic issue, this shows that a round trip may change the syntax and therefore the interpretation significantly (for instance, an addtional newline after **sttg** renders the following **payl** as though it were a **vtta** (making it disappear in playback)


src/media_tools/webvtt.c
```diff
@@ -430,7 +443,7 @@ GF_Err wvtt_dump(GF_Box *a, FILE * trace)
 	GF_WebVTTSampleEntryBox *cuebox = (GF_WebVTTSampleEntryBox *)a;
 	gf_isom_box_dump_start(a, "WebVTTSampleEntryBox", trace);
 	fprintf(trace, ">\n");
-	if (cuebox->config) boxstring_dump((GF_Box *)cuebox->config, trace);
+	//if (cuebox->config) boxstring_dump((GF_Box *)cuebox->config, trace);
```
jeanlf: is this line commented out because the box is dumped twice ? If so just remove the line

jeanlf:
thanks for the patch !

src/media_tools/webvtt.c
porcelijn: You're right, absolutely.
I forgot about that while tampering to figure out where the duplication came from. Removed it in dbe7aa6.

porcelijn:
Cheers, santé!

src/media_tools/webvtt.c
```diff
@@ -430,7 +443,6 @@ GF_Err wvtt_dump(GF_Box *a, FILE * trace)
 	GF_WebVTTSampleEntryBox *cuebox = (GF_WebVTTSampleEntryBox *)a;
 	gf_isom_box_dump_start(a, "WebVTTSampleEntryBox", trace);
 	fprintf(trace, ">\n");
-	if (cuebox->config) boxstring_dump((GF_Box *)cuebox->config, trace);
```
cconcolato: Why did you remove this line ? Is the box already dumped somewhere else?

porcelijn: Yes. I believe `gf_isom_box_dump_done()` prints it by default. I.e.: I saw MP4Box -diso report something like:
```
                <WebVTTConfigurationBox Size="19" Type="vttC" Specification="p30" Container="wvtt"><![CDATA[
WEBVTT FILE
]]></WebVTTConfigurationBox>
                <WebVTTConfigurationBox Size="19" Type="vttC" Specification="p30" Container="wvtt"><![CDATA[
WEBVTT FILE
]]></WebVTTConfigurationBox>
              </WebVTTSampleEntryBox>

```
even though there's obviously only one vttC. I wasn't sure if removal was the appropriate solution, but it results in a single vttC being written

porcelijn: @jeanlf also removed the local variable (`cuebox`) at line 430 which is no longer needed

cconcolato: Ok. Indeed, there was a change to preserve the order of boxes from the input to the dump, and for that change, we had to add all boxes in "other_boxes" instead of individual boxes. That's why we don't need to dump the 'config' here. So everything's fine. Thank you.

PR #872, opened 2017-07-13
Rfc 6381 size

rbouqueau:


rbouqueau:
Please merge when ok for you :)

jeanlf:
any reason why using 40 bytes max to store the codecs param ? This might be a bit short for LHVC content

rbouqueau:
Arbitrary value that looked sufficient to me. Which max would you prefer?

I could also store the max size in a central place. Which one? constants.h?

jeanlf:
yes, using a define in constant.h seems right

rbouqueau:
Which value >40?

jeanlf:
you can put 80 or 100 should be enough for now

rbouqueau:
better now? :)

jeanlf:
perfect, merged !

PR #886, opened 2017-07-31
Correct text in help main.c

Jamaika1:


aureliendavid:
the diff seems unreadable, it looks like most newline characters were removed

what did you change? can you resubmit your PRs with proper formatting?

thanks

PR #917, opened 2017-10-10
Add draft for subset of tests execution

DerouineauNicolas:
In some cases, it can be useful to make tests on a subset of gpac features.

Tested with:

./make_tests.sh -v scripts/dash-*
./make_tests.sh -v scripts/mp4client*

I'm not fluent in bash, so any comments or critics are welcome.


aureliendavid:
Added in 52595280b with the suggested syntax. Thanks.

Issue #883, opened 2017-07-23
MP4Box doesn't work well with non ASCII path: the remove file feature doesn't work


CyberSinh:
If I type the following command:

`"C:\Program Files\GPAC\mp4box.exe" -par 1=1:1 "E:\NAS\video\Vidéos de famille\2010\Alex 2010\HDV_0091.MP4"`

in a command prompt (cmd.exe) under Windows 10, the temp file "out_HDV_0091.MP4" is correclty created, but the command cannot finish because there is an error when the soft try to remove the old file:

```
Saving E:\NAS\video\Vid├®os de famille\2010\Alex 2010\HDV_0091.MP4: 0.500 secs Interleaving
Error removing file E:\NAS\video\Vid├®os de famille\2010\Alex 2010\HDV_0091.MP4
Error: I/O Error
```

I get the same error with Powershell:

```
Saving E:\NAS\video\VidÃ©os de famille\2010\Alex 2010\HDV_0091.MP4: 0.500 secs Interleaving
Error removing file E:\NAS\video\VidÃ©os de famille\2010\Alex 2010\HDV_0091.MP4
Error: I/O Error
```

The issue seems to be related with the space and the 'é' char contained in path. Perhaps the remove function called to drop the file doesn't support Unicode path.

As workaround, I use the "-out" parameter to not use the in-place rewrite feature. I manually delete old video files after processing.

rbouqueau:
Windows 10 here, and it seems to work in 'cmd':
```
C:\Users\rbouqueau>MP4Box -par 1=1:1 -add "D:\Works\gpac\issues\883_unicode\Vidéos de famille\counter_video_180.mp4" -new "D:\Works\gpac\issues\883_unicode\Vidéos de famille\tmp.mp4"
IsoMedia import counter_video_180.mp4 - track ID 1 - Video (size 320 x 180)
Saving D:\Works\gpac\issues\883_unicode\Vid├®os de famille\tmp.mp4: 0.500 secs Interleaving

C:\Users\rbouqueau>
```

Are you using the latest version of GPAC? What other information could you provide us so that we can reproduce?

CyberSinh:
I use the latest stable version (x64) on Windows 10.

The 'new' parameter was not used.
The input file was not on the same partition than the temp one.

rbouqueau:
That's not the first time users report this issue @jeanlf @aureliendavid  . We usually tell to specify the ```-tmp``` dir manually but may need a more automatic solution.

[...]

PR #919, opened 2017-10-17
Use widechar crt/api on Win32

sorayuki:
I see there are some commits which introduce wide char handling on Win32.
It's very helpful for CJK language users.
But I think some code are incorrect.
"char*" in Windows API / C runtime on Windows often refers to string in local 
encoding, not UTF8 string. Using wmain on Windows to receive wide char 
argument is a good idea, but it's not lossless to convert wide char string to 
local string.

I modified the code to convert argument string to utf8 on Windows,
and it uses utf8 in internal program logic, and wide char string for file api 
on Windows.

Note: 
If console code page is not set to utf8, the output log will still be mojibake.

### DIFF
--- a/applications/mp4box/main.c
+++ b/applications/mp4box/main.c
@@ -5279,39 +5279,25 @@ exit:
 }
 
 #if defined(WIN32) && !defined(NO_WMAIN)
-#include <windows.h>
-
-char* wstr_to_windows_multibyte(wchar_t* wstr) {
-
-	int size_needed;
-	char* res;
-
-	if (!wstr)
-		return NULL;
-
-	size_needed = WideCharToMultiByte(GetACP(), 0, wstr, -1, NULL, 0, NULL, NULL);
-
-	res = (char *)malloc(size_needed*sizeof(char));
-
-	WideCharToMultiByte(GetACP(), 0, wstr, -1, res, size_needed, NULL, NULL);
-
-	return res;
-
-
-}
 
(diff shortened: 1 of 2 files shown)


aureliendavid:
This looks like a more complete solution to what I was trying to do with 1516ac11e62666fa7ce285f6

Here was my rationale for doing it that way: https://github.com/gpac/gpac/issues/883#issuecomment-320915823

I was being lazy and trying to get dispay and file operations to work properly with minimum code change. 

I'll run some tests before merging but it looks like using wchars for file operations is the only way to support all characters.

aureliendavid:
tests look good

display will still be a problem since we use fprintf, but console charset support is a horrible mess on windows

For a test file with name: `bloopéàCJK乕伫aa.mp4`
With the PR, it will be opened/moved properly. 
By default it will display as: `bloop├⌐├áCJKΣ╣òΣ╝½aa.mp4`
By setting console codepage to UTF-8 it displays as: `bloopéàCJK???aa.mp4`

So even with the right codepage Windows can't really dispaly all utf8 characters, but it is still better than the default. 

It is possible to force setting the console codepage for the current process with a call to `SetConsoleOutputCP(65001);`  in the `wmain` (for example before calling `mp4boxMain()`).


In any case it's more important that the file operations work with everything, even if the display is funky. 

I'll merge the PR later today. Thanks @sorayuki for taking the time to implement it.

PR #1002, opened 2018-02-28
TSI/TOI values should be parsed as xsd:unsignedInt

sdevine01:
I recommend treating these two XML attributes as unsigned integers.  I believe a change to atoll() will do this.


jeanlf:
Thanks for the report. This was actually fixed on a dev branch, now merged.

PR #1042, opened 2018-05-01
add segmentation option to insert a tfdt per traf

cconcolato:
Useful when muxed segmented streams are processed separately by the client (e.g. like two independent CMAF tracks sharing only the moov)


rbouqueau:
Thank you. LGTM except for the formatting.

PR #1043, opened 2018-05-01
fix segfault on QT audio files

cconcolato:
MP4Box segfaults when parsing some QT files. The data array is null, possibly already parsed. This is a proposed fix that solves the problem but any other fix would work.


rbouqueau:
LGTM except for the formatting.

PR #1045, opened 2018-05-02
ffmpeg 4.0

psiniemi:
Hi, I don't know if you are interested, but this is what I had to do to get 0.7.1 to compile against ffmpeg 4.0 that removed some deprecated flags.


jeanlf:
could you do the PR without touching the ffmpeg include files (copied for android / win builds, should not be modified) ?

psiniemi:
Cleaned up in https://github.com/gpac/gpac/pull/1063

PR #1060, opened 2018-05-21
fix wrong log message

cconcolato:
Simple fix. When importing an NHML with a Sample with mediaFile and that file was missing, the file not found message was not the right one.


rbouqueau:
Merged. Thank you!

PR #1065, opened 2018-05-22
Add 'sdtp' as child of 'traf'

### LINKED ISSUE #1064: MP4Box issues warning when 'sdtp' box is present in 'traf'
The current code defines the 'sdtp' box as a child of 'stbl' only:
https://github.com/gpac/gpac/blob/71effb905f13152cb29406fa84eb12b599e1e3dd/src/isomedia/box_funcs.c#L845
In PIFF files, 'sdtp' is allowed in 'traf', according to https://go.microsoft.com/?linkid=9682897. We should allow that.

cconcolato:
Closes #1064


cconcolato:
@jeanlf is there a problem with this request?

jeanlf:
I'd rather keep the warning that it should not be there

PR #1070, opened 2018-05-31
Added composing of the mfra box when producing fragmented mp4.

ashpekvs:
mfra is created according to ISO/IEC 14496-12:2015(E)
part 8.8.9 Movie Fragment Random Access Box
Presence of this box speeds up seeking over fragmented file

It also fixes some issues on mfra write out (eg missed writing box header, etc)
Debugging/testing were done @win32 using msvc compiler on gpac_mp4box_mini solution.

As a proposal maybe it makes sense to add new cmd line option for this mfra box? (say "-mfra)


jeanlf:
thanks, I fixed a few things and moved the code in isomedia , could you check the latest git works fine for your test cases ?

jeanlf:
done thanks

PR #1082, opened 2018-06-07
Add AV1 support

cconcolato:
Creating PR to ease the review
Should NOT be merged now.

TODO list (post-added by @rbouqueau):
 - [x] add MP4Box option to force the high level syntax parsing e.g. ```"-add file.*:fmt=AnnexB"```
 - [x] https://github.com/AOMediaCodec/av1-isobmff/pull/33/files
 - [x] round-tripping MP4->OBU->MP4 does not seem to work. One problem is the output file is called file.av01 but the import only takes file.av1. Another one is the missing temporal delimiters.
 - [x] AnnexB import should add the obu_size again when absent
 - [ ] write unit tests
 - [ ] compute presentation_delay whenever possible: https://github.com/AOMediaCodec/av1-isobmff/issues/31
 - [ ] implement sample_groups features + see https://github.com/gpac/gpac/pull/1082#discussion_r193924724 (most important is forward key frames)
 - [x] check to remove ```assert(obu_size == gf_bs_get_position(bs) - pos);``` => doesn't assert after JLF's commit
 - [ ] DISABLE_AV1 should disable more code
 - [ ] move av1 specific code in separate file
 - [ ] retrieve sample timings from IVF frame header, see https://github.com/gpac/gpac/pull/1082#discussion_r193921757
 - [ ] detect WebM (not supported in GPAC)
 - [ ] ```Intra-only frames SHOULD be signaled using the sample_depends_on flag set to 2.```
 - [x] add file extensions: *.ivf or *.obu files

Last updated on GPAC side on libaom 265d15d46455ab4e208d3f9baaa840e05bd025fc (May 31st 2018).

### DIFF
--- a/src/isomedia/avc_ext.c
+++ b/src/isomedia/avc_ext.c
@@ -2503,6 +2503,7 @@ GF_Err av1c_Read(GF_Box *s, GF_BitStream *bs) {
 		GF_AV1_OBUArrayEntry *a;
 		
 		pos = gf_bs_get_position(bs);
+		obu_size = 0;
 		if (gf_media_aom_av1_parse_obu(bs, &obu_type, &obu_size, &state) != GF_OK) {
 			GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("[ISOBMFF] could not parse AV1 OBU at position "LLU". Leaving parsing.\n", pos));
 			break;
@@ -2538,8 +2539,7 @@ GF_Err av1c_Write(GF_Box *s, GF_BitStream *bs) {
 	GF_AV1ConfigurationBox *ptr = (GF_AV1ConfigurationBox*)s;
 	if (!s) return GF_BAD_PARAM;
 	if (!ptr->config) return GF_BAD_PARAM;
-
-	e = gf_isom_box_write_header(s, bs);
+	e = gf_isom_full_box_write(s, bs);
 	if (e) return e;
 
 	return gf_odf_av1_cfg_write_bs(ptr->config, bs);
--- a/src/isomedia/avc_ext.c
+++ b/src/isomedia/avc_ext.c
@@ -1546,6 +1546,8 @@ static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg) {
 	for (i=0; i<gf_list_count(cfg->obu_array); ++i) {
 		GF_AV1_OBUArrayEntry *dst = gf_malloc(sizeof(GF_AV1_OBUArrayEntry)), *src = gf_list_get(cfg->obu_array, i);
 		dst->obu_length = src->obu_length;
+		dst->obu_type = src->obu_type;
+		dst->obu = gf_malloc(dst->obu_length);
 		memcpy(dst->obu, src->obu, src->obu_length);
 		gf_list_add(out->obu_array, dst);
 	}
@@ -2407,8 +2409,6 @@ void av1c_del(GF_Box *s) {
 	gf_free(ptr);
 }
 
-GF_Err gf_media_aom_av1_parse_obu(GF_BitStream *bs, u64 *obu_size, ObuType *obu_type, AV1State *state);
-
 GF_Err av1c_Read(GF_Box *s, GF_BitStream *bs) {
 	AV1State state;
 	u8 reserved;
@@ -2429,10 +2429,11 @@ GF_Err av1c_Read(GF_Box *s, GF_BitStream *bs) {
 		GF_AV1_OBUArrayEntry *a;
 		
 		pos = gf_bs_get_position(bs);
-		if (gf_media_aom_av1_parse_obu(bs, &obu_size, &obu_type, &state) != GF_OK) {
+		if (gf_media_aom_av1_parse_obu(bs, &obu_type, &state) != GF_OK) {
 			GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("ISOBMFF: could not parse AV1 OBU at position "LLU". Leaving parsing.\n", pos));
 			break;
 		}
+		obu_size = gf_bs_get_position(bs) - pos;
 		gf_bs_seek(bs, pos);
 		GF_LOG(GF_LOG_DEBUG, GF_LOG_CONTAINER, ("ISOBMFF: parsed AV1 OBU type=%u size="LLU" at position "LLU".\n", pos));
 
@@ -2442,6 +2443,7 @@ GF_Err av1c_Read(GF_Box *s, GF_BitStream *bs) {
 		GF_SAFEALLOC(a, GF_AV1_OBUArrayEntry);
 		gf_bs_read_data(bs, a->obu, (u32)obu_size);
 		a->obu_length = obu_size;
+		a->obu_type = obu_type;
 		gf_list_add(ptr->config->obu_array, a);
 	}
 
--- a/src/isomedia/avc_ext.c
+++ b/src/isomedia/avc_ext.c
@@ -1196,6 +1196,64 @@ GF_Err AVC_HEVC_UpdateESD(GF_MPEGVisualSampleEntryBox *avc, GF_ESD *esd)
 }
 
 
+
+static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg) {
+	u32 i = 0;
+	GF_AV1Config *out = gf_malloc(sizeof(GF_AV1Config));
+	out->initial_presentation_delay_present = cfg->initial_presentation_delay_present;
+	out->initial_presentation_delay_minus_one = cfg->initial_presentation_delay_minus_one;
+	out->obu_array = gf_list_new();
+	for (i = 0; i<gf_list_count(cfg->obu_array); ++i) {
+		GF_AV1_OBUArrayEntry *dst = gf_malloc(sizeof(GF_AV1_OBUArrayEntry)), *src = gf_list_get(cfg->obu_array, i);
+		dst->obu_length = src->obu_length;
+		dst->obu_type = src->obu_type;
+		dst->obu = gf_malloc((size_t)dst->obu_length);
+		memcpy(dst->obu, src->obu, (size_t)src->obu_length);
+		gf_list_add(out->obu_array, dst);
+	}
+	return out;
+}
+
+void AV1_RewriteESDescriptorEx(GF_MPEGVisualSampleEntryBox *av1, GF_MediaBox *mdia)
+{
+	GF_BitRateBox *btrt = gf_isom_sample_entry_get_bitrate((GF_SampleEntryBox *)av1, GF_FALSE);
+
+	if (av1->emul_esd) gf_odf_desc_del((GF_Descriptor *)av1->emul_esd);
+	av1->emul_esd = gf_odf_desc_esd_new(2);
+	av1->emul_esd->decoderConfig->streamType = GF_STREAM_VISUAL;
+	av1->emul_esd->decoderConfig->objectTypeIndication = GPAC_OTI_VIDEO_AV1;
+
+	if (btrt) {
+		av1->emul_esd->decoderConfig->bufferSizeDB = btrt->bufferSizeDB;
+		av1->emul_esd->decoderConfig->avgBitrate = btrt->avgBitrate;
+		av1->emul_esd->decoderConfig->maxBitrate = btrt->maxBitrate;
+	}
+	if (av1->descr) {
+		GF_Descriptor *desc, *clone;
+		u32 i = 0;
+		while ((desc = (GF_Descriptor *)gf_list_enum(av1->descr->descriptors, &i))) {
+			clone = NULL;
+			gf_odf_desc_copy(desc, &clone);
+			if (gf_odf_desc_add_desc((GF_Descriptor *)av1->emul_esd, clone) != GF_OK)
+				gf_odf_desc_del(clone);
+		}
+	}
+
+	if (av1->av1_config) {
+		GF_AV1Config *av1_cfg = AV1_DuplicateConfig(av1->av1_config->config);
+		if (av1_cfg) {
+			gf_odf_av1_cfg_write(av1_cfg, &av1->emul_esd->decoderConfig->decoderSpecificInfo->data, &av1->emul_esd->decoderConfig->decoderSpecificInfo->dataLength);
+			gf_odf_av1_cfg_del(av1_cfg);
+		}
+	}
+}
+
+void AV1_RewriteESDescriptor(GF_MPEGVisualSampleEntryBox *av1)
+{
+	AV1_RewriteESDescriptorEx(av1, NULL);
+}
+
+
 #ifndef GPAC_DISABLE_ISOM_WRITE
 GF_EXPORT
 GF_Err gf_isom_avc_config_new(GF_ISOFile *the_file, u32 trackNumber, GF_AVCConfig *cfg, char *URLname, char *URNname, u32 *outDescriptionIndex)
@@ -1536,22 +1594,6 @@ GF_Err gf_isom_hevc_config_new(GF_ISOFile *the_file, u32 trackNumber, GF_HEVCCon
 	*outDescriptionIndex = gf_list_count(trak->Media->information->sampleTable->SampleDescription->other_boxes);
 	HEVC_RewriteESDescriptor(entry);
 	return e;
-}
-
-static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg) {
-	u32 i = 0;
-	GF_AV1Config *out = gf_malloc(sizeof(GF_AV1Config));
-	out->initial_presentation_delay_minus_one = cfg->initial_presentation_delay_minus_one;;
-	out->obu_array = gf_list_new();
-	for (i=0; i<gf_list_count(cfg->obu_array); ++i) {
-		GF_AV1_OBUArrayEntry *dst = gf_malloc(sizeof(GF_AV1_OBUArrayEntry)), *src = gf_list_get(cfg->obu_array, i);
-		dst->obu_length = src->obu_length;
-		dst->obu_type = src->obu_type;
-		dst->obu = gf_malloc(dst->obu_length);
-		memcpy(dst->obu, src->obu, src->obu_length);
-		gf_list_add(out->obu_array, dst);
-	}
-	return out;
 }
 
 GF_EXPORT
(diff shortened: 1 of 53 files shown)


src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+{
+	assert(frame_size);
+	*frame_size = gf_bs_read_u32_le(bs);
+	if (*frame_size > 256 * 1024 * 1024) {
+		GF_LOG(GF_LOG_ERROR, GF_LOG_CODING, ("[IVF] Wrong frame size %u\n", *frame_size));
+		*frame_size = 0;
+		return GF_NON_COMPLIANT_BITSTREAM;
+	}
+
+	gf_bs_read_u64(bs); //skip
```
cconcolato: Could be used to get the timestamps in case of Variable Frame Rate, but probably ok to skip for now.

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+	return GF_OK;
+}
+
+static GF_Err av1_parse_obu_header(GF_BitStream *bs, ObuType *obu_type, Bool *obu_extension_flag, Bool *obu_has_size_field, u8 *temporal_id, u8 *spatial_id)
+{
+	Bool forbidden = gf_bs_read_int(bs, 1);
+	if (forbidden) {
+		return GF_NON_COMPLIANT_BITSTREAM;
+	}
+	assert(obu_type && obu_extension_flag && obu_has_size_field && temporal_id && spatial_id);
```
cconcolato: Why? This will certainly fail for typical bitstream (without extension flag for example).

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+	switch (obu_type) {
+	case OBU_SEQUENCE_HEADER:
+	case OBU_METADATA:
+		return GF_TRUE;
+	default:
+		return GF_FALSE;
+	}
+}
+
+static Bool av1_is_obu_frame(ObuType obu_type) {
```
cconcolato: Is this meant to ignore OBUs that don't go in an MP4 sample?

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+		}
+	}
+
+	if (opt_Leb128Bytes) {
+		*opt_Leb128Bytes = Leb128Bytes;
+	}
+	return value;
+}
+
+#define OBU_COPY \
```
cconcolato: Probably worth having a check on obu_length, no?

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+	a->obu = gf_malloc((size_t)obu_length); \
+	gf_bs_read_data(bs, a->obu, (u32)obu_length); \
+	a->obu_length = obu_length; \
+	a->obu_type = obu_type;
+
+static void av1_populate_state_from_obu(GF_BitStream *bs, u64 pos, u64 obu_length, ObuType obu_type, AV1State *state)
+{
+	if (av1_is_obu_header(obu_type)) {
+		OBU_COPY;
+		if (!state->frame_state.header_obus) state->frame_state.header_obus = gf_list_new();
```
cconcolato: could probably be a function for this "if" and the next one.

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+		gf_list_add(state->frame_state.header_obus, a);
+	}
+	if (av1_is_obu_frame(obu_type)) {
+		OBU_COPY;
+		if (!state->frame_state.frame_obus) state->frame_state.frame_obus = gf_list_new();
+		gf_list_add(state->frame_state.frame_obus, a);
+	}
+}
+
+GF_Err aom_av1_parse_obu_from_section5(GF_BitStream *bs, AV1State *state) {
```
cconcolato: The name is a bit misleading as it parses all OBUs until if finds a temporal delimiter. So it parses a Temporal Unit. I'd suggest renaming it: `aom_av1_parse_tu_from_section5_obus`

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+		assert(obu_length == gf_bs_get_position(bs) - pos);
+		
+		GF_LOG(GF_LOG_DEBUG, GF_LOG_CONTAINER, ("[AV1] Section5 OBU detected (size "LLU")\n", obu_length));
+		av1_populate_state_from_obu(bs, pos, obu_length, obu_type, state);
+	}
+
+	return GF_OK;
+}
+
+GF_Err aom_av1_parse_obu_from_annexb(GF_BitStream *bs, AV1State *state)
```
cconcolato: Same thing. I would rename it to `aom_av1_parse_tu_from_annexb`

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+			u64 obu_length = leb128(bs, &Leb128Bytes);
+			if (frame_unit_size < Leb128Bytes + obu_length) {
+				GF_LOG(GF_LOG_ERROR, GF_LOG_CODING, ("[AV1] Annex B frame_unit_size("LLU") < Leb128Bytes("LLU") + obu_length("LLU")\n", frame_unit_size, Leb128Bytes, obu_length));
+				return GF_NON_COMPLIANT_BITSTREAM;
+			}
+			GF_LOG(GF_LOG_DEBUG, GF_LOG_CONTAINER, ("[AV1] Annex B        OBU detected (size "LLU")\n", obu_length));
+			u64 pos = gf_bs_get_position(bs);
+			frame_unit_size -= Leb128Bytes;
+
+			e = gf_media_aom_av1_parse_obu(bs, &obu_type, &obu_length, state);
```
cconcolato: If a new `obu_inner_length` is passed to `gf_media_aom_av1_parse_obu`, a check on whether the values inside the OBU header (if any) and outside (in the Annex B) are the same could be added.

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+	if (state->frame_id_numbers_present_flag) {
+		//idLen = (additional_frame_id_length_minus_1 + delta_frame_id_length_minus_2 + 3);
+	}
+	//allFrames = (1 << NUM_REF_FRAMES) - 1;
+	if (state->reduced_still_picture_header) {
+		frame_state->key_frame = GF_TRUE;
+	} else {
+		Bool show_existing_frame = gf_bs_read_int(bs, 1);
+		if (show_existing_frame == GF_TRUE) {
+			/*frame_to_show_map_idx	f(3)
```
cconcolato: Parsing frame_to_show_map_idx is necessary to generate the https://aomediacodec.github.io/av1-isobmff/#forwardkeyframesamplegroupentry

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+			if (frame_type == KEY_FRAME) {
+				refresh_frame_flags = allFrames
+			}
+			if (film_grain_params_present) {
+				load_grain_params(frame_to_show_map_idx)
+			}*/
+			return;
+		}
+		AV1FrameType frame_type = gf_bs_read_int(bs, 2);
+		frame_state->key_frame = gf_bs_read_int(bs, 1)/*show_frame*/ && frame_type == KEY_FRAME && frame_state->seen_frame_header;
```
cconcolato: Could we split that on two lines for readability: reading the show_frame flag and then setting the key frame state? 
Also,  you set key_frame to true without checking the presence of sequence header in the same temporal unit?

src/media_tools/av_parsers.c
```diff
@@ -1545,6 +1545,595 @@ GF_Err gf_m4a_write_config(GF_M4ADecSpecInfo *cfg, char **dsi, u32 *dsi_size)
...
+		if (*obu_size >= 1 + obu_extension_flag)
+			*obu_size = *obu_size - 1 - obu_extension_flag;
+		else {
+			GF_LOG(GF_LOG_WARNING, GF_LOG_CODING, ("[AV1] computed OBU size "LLD" (input value = "LLU"). Skipping.\n", *obu_size - 1 - obu_extension_flag, *obu_size));
+			return GF_NON_COMPLIANT_BITSTREAM;
+		}
+	}
+	*obu_size += (gf_bs_get_position(bs) - pos);
+
+	if (*obu_type != OBU_SEQUENCE_HEADER && *obu_type != OBU_TEMPORAL_DELIMITER &&
```
cconcolato: This is assuming that only 1 operating point is parsed. If you have 2 operating points you will have to parse twice the bitstream. It might be best to have operating point specific states and to store the OBU there.

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+	}
+	gf_isom_set_track_enabled(import->dest, track_num, 1);
+	if (import->esd && !import->esd->ESID) import->esd->ESID = gf_isom_get_track_id(import->dest, track_num);
+	import->final_trackID = gf_isom_get_track_id(import->dest, track_num);
+	if (import->esd && import->esd->dependsOnESID) {
+		gf_isom_set_track_reference(import->dest, track_num, GF_ISOM_REF_DECODE, import->esd->dependsOnESID);
+	}
+	gf_isom_set_cts_packing(import->dest, track_num, GF_TRUE);
+
+	e = gf_media_aom_parse_ivf_file_header(bs, &state);
```
cconcolato: Sniffing the type is probably fine for IVF because of the signature but I'm not sure about the AnnexB. Why don't you use the "-add file.av1:fmt=AnnexB" option? with the default being OBU.

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+			av1_bs_syntax = OBUs;
+		}
+	}
+
+	gf_bs_seek(bs, pos);
+	while (gf_bs_available(bs)) {
+		av1_reset_frame_state(&state.frame_state);
+		pos = gf_bs_get_position(bs);
+
+		/*we process each TU and extract only the necessary OBUs*/
```
cconcolato: Again all three methods should be called `...parse_tu_from...`

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+					GF_AV1_OBUArrayEntry *a = (GF_AV1_OBUArrayEntry*)gf_list_get(state.frame_state.header_obus, 0);
+					gf_list_add(av1_cfg->obu_array, a);
+					gf_list_rem(state.frame_state.header_obus, 0);
+				}
+
+				e = gf_isom_av1_config_new(import->dest, track_num, av1_cfg, NULL, NULL, &di);
+				if (e) goto exit;
+			} else {
+				/*safety check: we only support static metadata*/
+				if (gf_list_count(state.frame_state.header_obus) > gf_list_count(av1_cfg->obu_array)) {
```
cconcolato: Given that you already iterate through all frame_state.header_obus below, you don't need to do this check. Also technically, you could have repeated sequence header or metadata OBUs (weird but possible). I would remove this check.

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+					goto exit;
+				}
+				while (gf_list_count(state.frame_state.header_obus)) {
+					u32 i = 0;
+					GF_AV1_OBUArrayEntry *a_hdr = (GF_AV1_OBUArrayEntry*)gf_list_get(state.frame_state.header_obus, 0);
+					for (i = 0; i < gf_list_count(av1_cfg->obu_array); ++i) {
+						GF_AV1_OBUArrayEntry *a_cfg = (GF_AV1_OBUArrayEntry*)gf_list_get(av1_cfg->obu_array, i);
+						if (a_cfg->obu_type == a_hdr->obu_type) {
+							if (a_cfg->obu_length != a_hdr->obu_length || memcmp(a_cfg->obu, a_hdr->obu, (size_t)a_hdr->obu_length)) {
+								gf_import_message(import, GF_NOT_SUPPORTED, "Cannot find file %s", import->in_name);
```
cconcolato: Change the log message :)

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+			gf_list_rem(frame_state->frame_obus, 0);
+			gf_free(a);
+		}
+		gf_list_del(frame_state->frame_obus);
+	}
+
+	memset(frame_state, 0, sizeof(AV1StateFrame));
+}
+
+static GF_Err gf_import_aom_av1(GF_MediaImporter *import)
```
cconcolato: There does not seem to be an option of sample entry changes, which would be required if max_width/max_height changes.

cconcolato:
Should also update `MP4Box -h format`

src/isomedia/avc_ext.c
```diff
@@ -2341,6 +2457,113 @@ GF_Err hvcc_Size(GF_Box *s)
...
+		u64 pos, obu_size;
+		ObuType obu_type;
+		GF_AV1_OBUArrayEntry *a;
+		
+		pos = gf_bs_get_position(bs);
+		if (gf_media_aom_av1_parse_obu(bs, &obu_type, &obu_size, &state) != GF_OK) {
+			GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("[ISOBMFF] could not parse AV1 OBU at position "LLU". Leaving parsing.\n", pos));
+			break;
+		}
+		assert(obu_size == gf_bs_get_position(bs) - pos);
```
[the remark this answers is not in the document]
rbouqueau: I agree that this test is debatable. This is a cross-check that what we parse has the right length.

> The current AV1/ISOBMFF spec requires that the size field is present for OBUs in the config. 

Ok but if there are not present, should we skip the OBU or add the size field?

> Some OBUs may not be parseable if the obu_size is not given (e.g. private metadata). I don't see how the bistream position can be updated by gf_media_aom_av1_parse_obu in that case. This assert should be removed and replaced with the above.

It can't. In my opinion the AV1 spec should not allow that in any way.

src/media_tools/av_parsers.c
rbouqueau: Correct.

src/media_tools/av_parsers.c
rbouqueau: Which kind of check would you imagine?

src/media_tools/av_parsers.c
rbouqueau: > Could we split that on two lines for readability: reading the show_frame flag and then setting the key frame state?

Done.

> Also, you set key_frame to true without checking the presence of sequence header in the same temporal unit?

Correct. I skipped that check because the AV1 doesn't seem to prevent having the sequence header either 1) several times or 2) after the frame header. In this implementation we naively overwrite in the latest parsed struct.

I re-read the AV1 in ISOBMF document (```It contains a Sequence Header before the first Frame Header.```) and I think I was too conservative. So I've added the check.

src/media_tools/media_import.c
rbouqueau: I've added this in the TODO (see your first msg on this PR). This sniffing worked well on my fuzzing test.

src/media_tools/av_parsers.c
rbouqueau: Please teach me more on this :)

src/media_tools/media_import.c
rbouqueau: I'm not sure to understand. If the obus stored in the moov differ from the ones found in the first TU, then our current implementation fails. That's a limitation.

src/media_tools/media_import.c
rbouqueau: Thanks

rbouqueau:
I've pushed updates in 160dd25b13c6dc48ea1a6d154dbea510771be416 + added a TODO list at the top of this PR.

> Given how GPAC stores OBUs, the problem only happens if importing an AV1 stream from an MP4 generated by another tool that does not store the size field for the last OBU in the sample.
This could happen for very-low bitrate streams. In that case, most samples will be made of only one OBU (OBU_FRAME) and so storing the OBU length in the header could be important.

Indeed we would fail to import the last sample. I think no one will implement this optimization because it requires to rewrite the high level bitstream anyway.

src/media_tools/media_import.c
```diff
@@ -7018,6 +7016,237 @@ static GF_Err gf_import_hevc(GF_MediaImporter *import)
...
+					gf_list_add(av1_cfg->obu_array, a);
+					gf_list_rem(state.frame_state.header_obus, 0);
+				}
+
+				e = gf_isom_av1_config_new(import->dest, track_num, av1_cfg, NULL, NULL, &di);
+				if (e) goto exit;
+			} else {
+				/*safety check: we only support static metadata*/
+				if (gf_list_count(state.frame_state.header_obus) > gf_list_count(av1_cfg->obu_array)) {
+					gf_import_message(import, GF_NOT_SUPPORTED, "More header OBUs in than ", import->in_name);
```
cconcolato: Error message should be fixed.

cconcolato:
Also one thing I noted that could be improved is the support for file extensions. The reference encoder produces *.ivf or *.obu files. They should be supported.

cconcolato:
Also it'd be good if `-raw` would return an OBU sequence.

cconcolato:
More precisely, round-tripping MP4->OBU->MP4 does not seem to work. One problem is the output file is called `file.av01` but the import only takes `file.av1`. Another one is the missing temporal delimiters.

rbouqueau:
@jeanlf for me it is ready to merge

Issue #1080, opened 2018-06-06
MP4Box: write HEIF with ICC color profile


jonsneyers:
Is there currently any way to insert an ICC color profile when creating a HEIF/HEIC file with MP4Box? In the source code I see there is a function called `colr_Write()` which does seem to support writing a blob into a ColourInformationBox, which is what would be needed here. But I cannot figure out the syntax to MP4Box to make it actually insert a ColorInformationBox in the right spot, which would be in the ItemPropertyContainerBox of the metadata already written by MP4Box when you use `-add-image`. It feels like there might be some way to do an `-add-item` to add the color profile, but if there is, I can't figure out how to do it.

This is important since without this capability, we are basically stuck with sRGB, while most devices that support HEIC have P3-capable displays.

PR #1086, opened 2018-06-27
ICC color profile support for HEIF

jonsneyers:
This pull request closes issue #1080. It adds a command-line option to MP4Box (`icc_path`) to add an ICC color profile when writing a HEIF file. Also, when reading a HEIF file with an ICC color profile, it adds it (base64 encoded) to the IsoMedia XML dump.

This change is needed to correctly read and write HEIF images with a non-sRGB color profile, for example HEIF images created by iOS devices (which use the wider gamut P3 color profile).


jeanlf:
seems OK - any reason why you commented the check on AVC/HEVC/JPEG types ?

jonsneyers:
That check was preventing us from extracting the `grid` item type, which we need to reconstruct tiled HEIF images as created on iOS devices.

Issue #1114, opened 2018-08-15
Request for extended VP9 codec info in mpd manifest


BucherTomas:
When dashing, mp4box currently generates for VP9 video in mp4 container `vp09` codec info in mpd manifest. This is no longer supported on all recent browsers that can decode VP9. The codec string needs to include all the details such as profile, level, etc. as per https://www.webmproject.org/vp9/mp4/#codecs-parameter-string . 

This can be demonstrated by using this nifty webtool by one of your developers http://cconcolato.github.io/media-mime-support/ on Chrome or recent Firefox.

Existing `video/mp4;codecs="vp09"` on the page already says that `MediaSource.isTypeSupported = false`.

However, if you enter `vp09.00.50.08` (as generated for example from FFmpeg dash muxer for UHD content with level 5, 8-bit) into the testing box, then `MediaSource.isTypeSupported = true`

Default mpd manifest from mp4box with vp09 codec info is incompatible with any browser, once I manually edit the codec info to vp09.00.50.08, the codec negotiation is successful and the stream starts to play.

It would be great to have mp4box generate the codec info properly by itself as it does for avc1 or hevc.

aureliendavid:
(see #1115 for a follow up)

aureliendavid:
should be fixed, reopen if needed

BucherTomas:
@aureliendavid It looks like you are reading the vaules from vpcC box as hexadecimal, however, they are stored as decimal. The documentation version I referred to unfortunately lacks the emphasis, but for example https://github.com/webmproject/vp9-dash/blob/f8d543ef1006b9c5b4661141b822dde67d955283/VPCodecISOMediaFileFormatBinding.md states it understandably enough.

aureliendavid:
dammit you're right

looks like it's just for the `level` field, which of course is the only one that ffmpeg doesn't set itself so my test sequences were wrong (should be set with `-level 0x50` and not `-level 50` in ffmpeg) 

(as far as I understood the link, other fields are regular uints, for the bit depth for example, if it's 10bit it will be encoded a `0x0a` and not `0x10`)

there's always something, isn't there..

BucherTomas:
Thanks, now it looks good.

[...]

PR #1115, opened 2018-08-16
added parsing of vpcc box for VP8/9 (#1114)

aureliendavid:
[I'm putting this in a PR in case someone wants to review it before merging]

see #1114 for context

I added functions to parse the `vpcC` box of VP9 tracks base on this spec: https://www.webmproject.org/vp9/mp4

This allows to build an accurate codec info string for dash. I've probably missed some things but it should be enough for this particular purpose. 

[Note that when converting webm to mp4 with ffmpeg, the `level` field is lost and set to 0. I don't know if it's ffmpeg that doesn't copy it or if it's not actually stored in the webm file. In any case it works when giving it explicitly to ffmpeg: 

`ffmpeg -i in.web -c copy -level 50 out.mp4`

There's actually a commit at ffmpeg to guess the level based on other parameters but it's not in any release yet: https://github.com/FFmpeg/FFmpeg/commit/5b6cc3a73acb90df0d4fe5ca8f51527b64592bf8 ]

### DIFF
--- a/include/gpac/internal/isomedia_dev.h
+++ b/include/gpac/internal/isomedia_dev.h
@@ -185,6 +185,11 @@ enum
 	GF_ISOM_BOX_TYPE_AV1C = GF_4CC('a', 'v', '1', 'C'),
 	GF_ISOM_BOX_TYPE_AV01 = GF_4CC('a', 'v', '0', '1'),
 
+	/* WebM */
+	GF_ISOM_BOX_TYPE_VPCC = GF_4CC('v', 'p', 'c', 'C'),
+	GF_ISOM_BOX_TYPE_VP08 = GF_4CC('v', 'p', '0', '8'),
+	GF_ISOM_BOX_TYPE_VP09 = GF_4CC('v', 'p', '0', '9'),
+
 	/*LASeR extension*/
 	GF_ISOM_BOX_TYPE_LSRC	= GF_4CC( 'l', 's', 'r', 'C' ),
 	GF_ISOM_BOX_TYPE_LSR1	= GF_4CC( 'l', 's', 'r', '1' ),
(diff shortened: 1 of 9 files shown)


jamrial:
> [Note that when converting webm to mp4 with ffmpeg, the level field is lost and set to 0. I don't know if it's ffmpeg that doesn't copy it or if it's not actually stored in the webm file.

Level can be stored in Matroska/WebM in a custom struct stored in CodecPrivate as defined in https://www.webmproject.org/docs/container/
ffmpeg doesn't currently read or write it, and i haven't really seen any VP9 in WebM file with it, not even YouTube videos, so i'm not sure if it will ever see any real usage at all.

That being said, it's worth adding support for that VP9 specific CodecPrivate struct in case a file ever features it.

cconcolato:
I tested it on some files. It does seem to work but you might want to:
- allow the 'colr' box in that sample entry. It currently says:
>Box "colr" is invalid in container vp09

- support the 'SmDm' and 'CoLL' boxes
>[iso file] Unknown box type SmDm
>[iso file] Unknown box type CoLL

- Change the track type, it currently says "Unknown track type"
> Track # 1 Info - TrackID 1 - TimeScale 60000
Media Duration 00:00:00.000 - Indicated Duration 00:00:00.000
Track has 1 edit lists: track duration is 00:03:49.812
Media Info: Language "Undetermined (und)" - Type "vide:vp09" - 0 samples
Fragmented track: 13775 samples - Media Duration 00:03:49.812
Visual Track layout: x=0 y=0 width=1280 height=720
Unknown track type
	RFC6381 Codec Parameters: vp09.02.10.10.01.09.16.09.00
	All samples are sync

The "All samples are sync" is weird I have to investigate.

cconcolato:
Also note that for encrypted VP9 tracks, MP4Box reports "WARNING: Broken MPEG-4 Track".

PR #1123, opened 2018-08-21
add -mvex-after-traks option to MP4Box when dashing

cconcolato:
Some devices seem to have a problem when `mvex` is provided before `trak`. Also Table 3 in CMAF places `mvex` after, although order of boxes is not mandated in CMAF.


cconcolato:
Also note that HLS requires that, see https://tools.ietf.org/html/rfc8216#section-3.3

PR #1144, opened 2018-09-24
fix use of -no-frag-defaults

cconcolato:
Files generated today with -no-frag-defaults do not pass CMAF validation checks. These changes fix that problem. Could be also behind another flag if -no-frag-defaults is really used.


jeanlf:
looks OK, thanks for the fix

PR #1163, opened 2018-11-08
Fix incorrect variable check order in NewMedia function

ChenXiaomin36:
Change-Id: Iacc803a654935940c76a8a9de34a316102ae166e


jeanlf:
thanks!

PR #1172, opened 2018-11-29
add -dsapx to generate cues from the sidx

cconcolato:
It can be useful to generate a cues.xml file from an existing fragmented file, to tweak the cues times. This PR adds that. 

It would be good to add the sample number and dts and cts, but if the file is opened in dump mode, it does not seem possible to get the sample. @jeanlf can you check?


jeanlf:
Wouldn't it be simpler to remember which sample started a fragment, rather than inspecting the sidx ? This would work with fragmented files that don't have any sidx

jeanlf:
we can now get the info on which sample was a fragment start, added to dsap. If needed we could patch to also track which sample is the first of a segement, in case you need multiple fragments per segment

PR #1173, opened 2018-11-29
adding support for audio_roll

cconcolato:
For AAC or HE-AAC it may be interesting to signal how much pre-roll is needed, as documented for example in [Apple documentation](https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFAppenG/QTFFAppenG.html)

### DIFF
--- a/src/media_tools/media_import.c
+++ b/src/media_tools/media_import.c
@@ -1167,7 +1167,7 @@ GF_Err gf_import_aac_adts(GF_MediaImporter *import)
 	samp->DTS+=dts_inc;
 
 	cur_samp++;
-	if (import->audio_roll) {
+	if (import->audio_roll_change) {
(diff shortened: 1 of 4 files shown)


src/media_tools/media_import.c
```diff
@@ -2326,6 +2338,8 @@ GF_Err gf_import_isomedia(GF_MediaImporter *import)
 
 		gf_isom_copy_sample_info(import->dest, track, import->orig, track_in, i+1);
 
+		gf_isom_set_sample_roll_group(import->dest, track, i+1, import->audio_roll);
```
jeanlf: 1- could you also enclose this one in a if (import->audio_roll) { } ?
2- maybe for isobmff import we want to reset the roll to 0 if the source track already has roll info ?

cconcolato: 1 - done
2 - done.

PR #1185, opened 2018-12-20
sidx fix for ssix

cconcolato:
I'm not totally sure the fix works for all sidx cases but it does for the simple case (a single sidx).


jeanlf:
not doing it that way, but should now be fixed, thanks for the report

PR #1201, opened 2019-02-04
Improved HEIF/MIAF/AVIF support

cconcolato:
Few comments about this PR:
- it enables addressing comments in https://github.com/AOMediaCodec/av1-avif/pull/50
- it adds support for the `pixi` box for all image items
- it adds the `auxi` box when the `alpha` import string is given. Meant to be used as in `-add file.obu:hdlr=auxv:alpha`.  
- it fixes a bug with writing of `auxC`
- the `meta_find_prop` seems buggy. It searches for a property that as the same value as the given in the `searchprop` field but `searchprop` is initialized to 0 which might be a value for which a property is searched... I did not fix that here.
- We've reached the end of `GF_MediaImporter flags` space (32 flags defined). I had to use a separate field `is_alpha`. Some harmonization may be needed here or at least some guidance on when to use flags and when to use fields.


jeanlf:
on  `meta_find_prop`, the property passed is for search is not set to 0, the fields are toggled before/after the call so this should work fine

PR #1209, opened 2019-02-22
Piff senc dump fix

cconcolato:
This PR fixes one problem: when the initialization vector starts with 0, no PIFFSampleEncryptionEntry was dumped because of the use of strlen.

The PR also harmonizes the dump between the PIFF senc box and the senc box (adding sample number in dump). Ideally the code should not be duplicated.


rbouqueau:
LGTM

Issue #1216, opened 2019-03-16
gpac build failed.  undefined reference to `lzma_code'


kuh3h3:
18:17:11 AMF git .......................................... [Up-to-date]
18:17:11   Running git clone for gpac...
18:18:14 ┌ gpac git ................................. [Recently updated]
18:18:16 ├ Running configure...
18:19:11 ├ Running make...
Likely error:
../../bin/gcc/libgpac_static.a(base_encoding.o):base_encoding.c:(.text+0xde2): undefined reference to `lzma_code'
../../bin/gcc/libgpac_static.a(base_encoding.o):base_encoding.c:(.text+0xe24): undefined reference to `lzma_end'
../../bin/gcc/libgpac_static.a(base_encoding.o):base_encoding.c:(.text+0xf21): undefined reference to `lzma_stream_decoder'
../../bin/gcc/libgpac_static.a(base_encoding.o):base_encoding.c:(.text+0xfb5): undefined reference to `lzma_code'
collect2.exe: error: ld returned 1 exit status
make[2]: *** [Makefile:64: MP4Box.exe] Error 1
make[2]: Leaving directory '/build/gpac-git/build-64bit/applications/mp4box'
make[1]: *** [Makefile:55: apps] Error 2
make[1]: Leaving directory '/build/gpac-git/build-64bit/applications'
make: *** [Makefile:10: all] Error 2
make failed. Check C:/Users/reandro/source/repos/media-autobuild_suite/build/gpac-git/build-64bit/ab-suite.make.log
This is required for other packages, so this script will exit.
18:24:29   Creating diagnostics file...

All relevant logs have been anonymously uploaded to https://0x0.st/zXhm.zip
Copy and paste [logs.zip](https://0x0.st/zXhm.zip) in the GitHub issue.
Make sure the suite is up-to-date before reporting an issue. It might've been fixed already.
Try running the build again at a later time.

kuh3h3:
if add -llzma to gpac-git\build-64bit\applications\mp4box\Makefile,
build ok
but what make this error?

ifeq ($(STATICBUILD),yes)
##include static modules and other deps for libgpac
include ../../static.mak
LINKFLAGS+=-lgpac_static -llzma
LINKFLAGS+= $(GPAC_SH_FLAGS)
LINKFLAGS+=$(EXTRALIBS)
else
LINKFLAGS+=-lgpac
endif

kuh3h3:
[...]

PR #1217, opened 2019-03-17
fix lzma linking in msys/mingw/cygwin

schmidthubert:
lzma test won't pass otherwise in `#look for lzma` support in configure and therefore linking will fail.

Fixes https://github.com/gpac/gpac/issues/1216


schmidthubert:
Are you sure this TravisCI thing is correctly configured? Why is there Ubuntu running in that TravisCI if this is a patch for msys/mingw/cygwin? It seems TravisCI is clueless and doesn't "know" anything about msys/mingw/cygwin ...

jeanlf:
should now be fixed, thanks for the report

PR #1234, opened 2019-04-13
check sample description entry type (either UUID or codec_tag) are not NULL

rbouqueau:
because it would lead to isobmff files that aren't parsed correctly, see [debug.zip](https://github.com/gpac/gpac/files/3076094/debug.zip)

we could probably be able to provide more relevant error messages, could someone have a look?


rbouqueau:
Can we merge?

PR #1277, opened 2019-07-17
opus update

rbouqueau:


aureliendavid:
(on a déjà eu le soucis mais quand le nom de la branche contient des underscores le build debian fail parce qu'on met le nom de la branche dans la version et debian refuse les underscores dans les noms de version)

src/filters/mux_isom.c
```diff
@@ -2522,6 +2522,8 @@ static GF_Err mp4_mux_process_sample(GF_MP4MuxCtx *ctx, TrackWriter *tkw, GF_Fil
 	sap_type = gf_filter_pck_get_sap(pck);
 	if (sap_type==GF_FILTER_SAP_1)
 		tkw->sample.IsRAP = SAP_TYPE_1;
+	else if (sap_type == GF_FILTER_SAP_4)
+		tkw->sample.IsRAP = SAP_TYPE_4;
```
jeanlf: I would suggest that we don't assign to SAP4 here (done later in that function), but rather assign to SAP1 in the case of non-visual media, otherwise we might break syncSample on many implementations that expect it to be always set for non visual tracks

src/filters/dmx_ogg.c
```diff
@@ -699,18 +699,30 @@ GF_Err oggdmx_process(GF_Filter *filter)
 					else if (st->info.type==GF_CODECID_OPUS) {
 						block_size = gf_opus_check_frame(st->opus_parser, (char *) oggpacket.packet, oggpacket.bytes);
 						if (!block_size) continue;
+
+						if (!st->recomputed_ts) {
+							gf_filter_pid_set_property(st->opid, GF_PROP_PID_DELAY, &PROP_SINT((s32)-st->opus_parser->PreSkip));
+						}
```
jeanlf: are we always sure that PreSkip is less than one block size ? If not we might need to signal preskip on several samples

rbouqueau: PreSkip can be more than one block size. However the edit list applies once to all the samples. What did I miss?

jeanlf: once the first packet will be sent, st->recomputed_ts will no longer be null. I think you need to store the PreSkip length in the context and decrement by block_size at each packet

rbouqueau:
If you are ok with that, please merge.

I still have unexpected fails in my MSYS2 but they are unrelated to ogg&opus:
```
$ LD_LIBRARY_PATH=$PWD/../bin/gcc PATH=$PWD/../bin/gcc:$PATH  ./make_tests.sh  -precommit scripts/mp4box-io.sh
[...]
mp4box-io-enst_audio.flac: dts:HashSourceNotFound diso:HashSourceNotFound add:HashSourceNotFound MediaTime:Fail XMLDump:Fail MediaInfo:Fail MediaImport:Fail
mp4box-io-logo.jp2: dts:HashFail diso:HashFail add:HashFail
mp4box-io-counter_30s_audio.ac3: hint:HashFail diso:HashFail
Number of Tests 384 - 1605 subtests - Executed: 24 (6 %) - Cached: 360 (93 %)
Tests passed 378 (98 %) - 1605 sub-tests
Tests failed 1 (0 % of subtests)
Tests Unknown 5 (1 % of subtests)
Tests HASH total 1310 - fail 22 (1 % of subtests)
Generation done in 00:00:40:310
```

PR #1283, opened 2019-07-21
fix av1 double free

rbouqueau:
The mem mode of AV1 creates this weird pattern regarding the data ownership. Is it what we expect?


jeanlf:
this should have been fixed by ecdb884

PR #1306, opened 2019-09-06
Removing duplicates in fileimport.c in MP4Box

RodolpheFouquet:
- the duplicate doesn't exist on the filters branch, which served as a guideline for this fix
- some errors could be overwritten, they are now managed.

### DIFF
--- a/applications/mp4box/fileimport.c
+++ b/applications/mp4box/fileimport.c
@@ -858,9 +858,11 @@ GF_Err import_file(GF_ISOFile *dest, char *inName, u32 import_flags, Double forc
             if (((par_n>=-1) && (par_d>=-1)) || force_par) {
                 e = gf_media_change_par(import.dest, track, par_n, par_d, force_par);
             }
+			if (e) goto exit;
             if (has_clap) {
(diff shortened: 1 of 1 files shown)


jeanlf:
you need to patch the following functions as well:
```
gf_import_timed_text
gf_import_raw_unit
```
These functions currently don't set the number of tracks on import probing, your patch will break these importers

RodolpheFouquet:
Hey, thanks, I'll fix that then.

RodolpheFouquet:
Is it possible to have command line examples so I can try before & after the patch and verify I'm not breaking that specific nb_track == 0 case?

rbouqueau:
Jean tells me it is about chapters.

I can see the coverage in https://tests.gpac.io/testres/lcov/1644. It seems like it is more about languages (as szLan) is non-null but I couldn't extract the proper test. Idea anyone?

RodolpheFouquet:
Well, about the tests, the same number of tests are passing before and after my PR (when I run the tests).

rbouqueau:
Since this only affects master, I merge.

PR #1355, opened 2019-11-20
Fix shift exponent is too large in gf_bs_write_int

stanionascu:
When importing mp4's containing dvcc box, MP4Box will fail complaining that:
`utils/bitstream.c:552:9: runtime error: shift exponent 4294967168 is too large for 32-bit type 'unsigned int'`

It was hapenning due to 160 being passed into the `gf_bs_write_int` (from dvcc_Write) and shifting the value beyond the valid range.

The fix ensures that the value is correctly written, by draining the count of bits all the way up until it reaches an acceptable value.

### DIFF
--- a/src/isomedia/box_code_base.c
+++ b/src/isomedia/box_code_base.c
@@ -11513,7 +11513,11 @@ GF_Err dvcC_Write(GF_Box *s, GF_BitStream *bs)
 	gf_bs_write_int(bs, ptr->DOVIConfig.rpu_present_flag, 1);
 	gf_bs_write_int(bs, ptr->DOVIConfig.el_present_flag, 1);
 	gf_bs_write_int(bs, ptr->DOVIConfig.bl_present_flag, 1);
-	gf_bs_write_int(bs, 0, 5*32);
+	{
+		u32 data[5];
+		memset(data, 0, sizeof(data));
+		gf_bs_write_data(bs, (char*)data, sizeof(data));
+	}
 
 	return GF_OK;
 }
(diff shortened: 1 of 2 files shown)


stanionascu:
Can anyone have a look at this ?

rbouqueau:
Yes we will :)
Thanks for reporting and sorry for the delays.

rbouqueau:
LGTM. @jeanlf Should we add a log message in this case?

jeanlf:
Looks good, but I think we should rather throw throw an error if nBits is greater than 32, since gf_bs_write_long_int should be used instead. Note that dvcC in filters does not use have this issue.

rbouqueau:
By "throw an error" you mean ```GF_LOG(GF_LOG_ERROR, ...```, correct?

jeanlf:
yes

rbouqueau:
@stanionascu Can you add it, or should I duplicate your branch to add commits?

stanionascu:
@rbouqueau sure I can amend the commit.

stanionascu:
@rbouqueau @jeanlf  
> Looks good, but I think we should rather throw throw an error if nBits is greater than 32, since gf_bs_write_long_int should be used instead. Note that dvcC in filters does not use have this issue.
> @stanionascu Can you add it, or should I duplicate your branch to add commits?

i've pushed latest changes:
* error logging and draining to both `gf_bs_write_*_int` methods.
* use `gf_bs_write_data` in `dvcC_Write` as symmetric to `gf_bs_read_data` in `dvcC_Read`

rbouqueau:
LGTM

@Jean please merge if ok for you

rbouqueau:
@stanionascu Many thanks :)

jeanlf:
thanks!

PR #1359, opened 2019-11-22
Fix an error while executing dpkg-genbuildinfo

ledyba-z:
`dpkg-buildpackage -j10 -uc -us` fails with such an error:

```bash
% dpkg-genbuildinfo

dpkg-genbuildinfo: error: cannot fstat file ../gpac-dbgsym_0.8.0_amd64.ddeb: No such file or directory
dpkg-buildpackage: error: dpkg-genbuildinfo subprocess returned exit status 255
```

In this patch, this issue is resolved by set destdir to parent directory.

Please take a look!

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -103,7 +103,7 @@ binary-arch: build install
 	dh_shlibdeps
 	dh_gencontrol
 	dh_md5sums
-	dh_builddeb --destdir=.
+	dh_builddeb --destdir=..
 
 binary: binary-indep binary-arch
 .PHONY: build clean binary-indep binary-arch binary install


aureliendavid:
you can also build the package with `make deb` which should work as is

however you're right that `dpkg-buildpackage` needs the .deb files to be in the parent dir to work properly

this has an impact on some scripts on our end so i'll only merge this when i'll be able to change and test our scripts 

thanks for the contrib

ledyba-z:
Thank you for reviewing.

Oh sorry, I didn't know that target in Makefile!
Thanks for telling it.

I can make debian packages with that Makefile target, so I will close this PR.

PR #1366, opened 2019-12-15
Filters master diffs

rbouqueau:
I'm trying to avoid diffs between both branches for existing users.

Missing:
 - [ ] Diffs with sidx/ssix
 - [ ] Timescales when Dashing


jeanlf:
looks good to me, thanks.

Issue #1368, opened 2019-12-19
DASH MPD: characters are never escaped which leads to non-compliant XMLs


rayfoss:
I found a stray broken MPD... because the path to it has an ampersand and the program info element inserts the path unfiltered.

```
 <ProgramInformation moreInformationURL="http://gpac.io">
  <Title>dash/1 T&D Systems Overview/video.mpd generated by GPAC</Title>
 </ProgramInformation>
```

Workaround: `-mpd-title "TD PowerSkills Training Video"`

https://github.com/gpac/gpac/blob/451c1a87af5c99b98bb7d1b30812517d4c99e9f9/src/media_tools/dash_segmenter.c#L5639

rbouqueau:
Hi, thanks for reporting. This is indeed an important limitation. I'm going to rename the issue and I hope one of us can have a look before the next release.

aureliendavid:
I'll try to take a look at it

aureliendavid:
should be ok now, see the discussion in #1369

PR #1369, opened 2019-12-19
escape xml entities in mpd (#1368)

aureliendavid:
Hi, 

(I'm doing this as a PR while the buildbot is still out of commission and it might need further tests/discussion.) 

This is meant to fix (at least partially) issue #1368.

Since string replacements are a pain in C, I felt it easier to fprintf character-by-character and replace on the fly. 

Since most cases are of the form `fprintf(mpd, "<something>%s</something>", string)` I added a utility function `gf_xml_dump_string(file, before, str, after)` so that the syntax can stay close. 

I replaced the most obvious cases but it might be needed elsewhere (`egrep -n "fprintf\(mpd.*%s.*" src/media_tools/dash_segmenter.c ` helps to see other places with this pattern).

A test with something like:
```
$ MP4Box -dash 1000 -base-url "http://example.com/dash?p=1&s=2&q='abc'" -out "a&bc.mpd" -profile-ext "prof&ex<\"t\">" -cprt "™, ℠, ® & ©" colorbars.mp4 
```
now produces an xml-valid mpd file.


rbouqueau:
Thank you.

The core issue in #1368 comes from a string provided by the user.

This PR currently only works with ASCII, correct? If so any code table of multi-byte encoding wouldn't be supported.

For the code we emit we can escape but I don't think we ever need to. We only need to escape for user-provided strings. It seems like things that come from the command-line are unreliable (see https://stackoverflow.com/a/5596232/6028399) anyway.

What do you think?

aureliendavid:
> 
> This PR currently only works with ASCII, correct? If so any code table of multi-byte encoding wouldn't be supported.
>

It should also work with utf-8 because of the property that no ASCII byte ever appear in a multi-byte utf-8 sequence. In other words there is no danger of misinterpreting the middle of a multi-byte character as being an `&` for example, and multibytes will be written as-is by the default case in the switch. 

It's by design in utf-8, see the "Backward compatibility" paragraph in https://en.wikipedia.org/wiki/UTF-8#Description
 
> For the code we emit we can escape but I don't think we ever need to. We only need to escape for user-provided strings. It seems like things that come from the command-line are unreliable (see https://stackoverflow.com/a/5596232/6028399) anyway.
> 

The ones I've changed for now are all settable by the command-line. 

On windows we start by converting the wide chars to utf-8 (https://github.com/gpac/gpac/blob/master/applications/mp4box/main.c#L5933) so it should work in most cases. 

On unixes it should work with locales that are either single-byte (like latin1) or have the property mentioned above (like utf8).

rbouqueau:
Thanks! LGTM

PR #1397, opened 2020-01-21
Use the same variables leads to logic chaos

LeoKingLong:
File：dmx_dash.c  Method：dashdmx_process
This method use the same  loop variables "i" in three traverses at the same time , it will sometimes cause failure when play the video. Redeclare "i" in new loop can avoid some unexpected exception.


jeanlf:
Ouch indeed ! Thanks for the report - I changed the var name, it's safer

PR #1409, opened 2020-02-10
Improving error reporting and error handling for XML parsing.

mc0re:
It took some effort for me to get the XML right for decryption to work. Part of it was a lack of descriptive error messages. So I added them in a few places, where I had problems myself.


jeanlf:
thanks for the feedback - not doing it exactly that way to avoid alloc/free of error messages but kept the principles.

Issue #1417, opened 2020-02-20
Please do not hardcode -L/usr/lib


tpgxyz:
- [x] I looked for a similar issue and couldn't find any.
- [x] I tried with the latest version of GPAC. Installers available at http://gpac.io/downloads/gpac-nightly-builds/
- [x] I give enough information for contributors to reproduce my issue (meaningful title, github labels, platform and compiler, command-line ...). I can share files anonymously with this dropbox: https://www.mediafire.com/filedrop/filedrop_hosted.php?drop=eec9e058a9486fe4e99c33021481d9e1826ca9dbc242a6cfaab0fe95da5e5d95

Hi,

i've noticed that gpac hardcodes in many places -L/usr/lib which is very wrong is case of x86_64,aarch64, riscv64 and many more 64 bit platforms where libraries are placed in /lib64, /usr/lib64

These hardcodes produces linking errors:
```
[...]

PR #1418, opened 2020-02-20
Get rid of spaces behind commas in ifeq/ifneq statements

berolinux:
make (or at least some implementations of it, such as GNU make 4.3) take
the space as a part of the expression following the comma, resulting in
a matching of variables against things like "[space]/usr", "[space]yes"
etc.

Fixes a part of https://github.com/gpac/gpac/issues/1417


jeanlf:
done directly on new arch, thanks for the report

PR #1424, opened 2020-03-04
use -p=0 by default for MP4Box

aureliendavid:
[doing this as a PR to be sure we all agree with the change]

In order to get closer behavior between filters and master for MP4Box, this change makes `-p=0` (i.e.: don't use a config file on disk) the default **for MP4Box only** 

I also changed the meaning of `-p=` (empty string) to the previous default (i.e.: write GPAC.cfg in the default dir, not in profiles dir) (`-p=` (empty) used to be synonymous with `-p=0` but undocumented)


jeanlf:
looks good to me

PR #1458, opened 2020-04-07
Fixing issue 1456 - UUID reading.

mc0re:
UUID block should have version and flags.
Also adjusted its length to match.


jeanlf:
fixed in 26ba3fb72

PR #1484, opened 2020-05-13
Dolby vision

rbouqueau:
DolbyVision support:
 -  [X] RFC8361
 -  [X] DASH (ok)
 -  [X] Encryption (ok)
 -  [ ] "Dolby recommends including a separate adaptation set that refers to an SDR version of the same content as the Dolby Vision stream." => filters only: put an "alt_codecs" on AS and force a  duplication in the MPD serializer by patchin the AS ID when serializing
 - [ ] Raw import: multiple stream structure reversed-engineered. Clarification with new streams needed.
 - [ ] Add test


jeanlf:
looks good to me

PR #1523, opened 2020-06-23
Catpl tmcd extend dur

rbouqueau:


rbouqueau:
@jeanlf Could you review and merge pleased? I'm being pinged for this code to be integrated in master :)

applications/mp4box/fileimport.c
```diff
@@ -1955,7 +1955,7 @@ GF_Err cat_isomedia_file(GF_ISOFile *dest, char *fileName, u32 import_flags, GF_
 
 					segmentDuration += (u64) (s64) dur;
 					gf_isom_modify_edit(dest, dst_tk, j, segmentDuration, mediaTime, editMode);
-					break;
+					continue;
 				}
 			}
 		} else {
```
jeanlf, line 1961: why are you changing the break to continue ? That code is supposed to change only one edit list entry

applications/mp4box/fileimport.c
```diff
@@ -2013,6 +2013,23 @@ GF_Err cat_isomedia_file(GF_ISOFile *dest, char *fileName, u32 import_flags, GF_
 		gf_media_update_bitrate(dest, dst_tk);
 
 	}
+	for (i = gf_isom_get_track_count(orig); i < gf_isom_get_track_count(dest); ++i) {
+		if (gf_isom_get_media_type(dest, i+1) == GF_ISOM_MEDIA_TIMECODE) {
```
jeanlf, line 2017: I don't understand this loop, do you assume tracks will be in the same order ?

applications/mp4box/fileimport.c
rbouqueau: I misread this, I thought we were iterating on the tracks, not the edits. Reverting.

applications/mp4box/fileimport.c
rbouqueau: Correct.

rbouqueau: Actually this is a continuation of the previous loop:
```
for (i=0; i<nb_tracks; i++) {
```

where ```nb_tracks = gf_isom_get_track_count(orig);```.

rbouqueau: The idea is that if there is a tmcd track in dest that was not covered from orig, then we should extend the last sample duration to match the new total duration.

Does it make sense?

jeanlf: the idea makes sense, but scanning from i=gf_isom_get_track_count(orig) assumes that the TMCD track will always be in the extra declared tracks, which might not always be the case (-cat will find the right source track for each track to concatenate).I would do a full scan from i=0 to dest.nb_tracks

rbouqueau: do you propose to simply iterate from zero then?

in this case we have to accept that we will align all tmcd tracks

jeanlf: arg right. So either we take the risk of missing one track, or we align all tmcd... Is there a use case (for now) for not aligning all tmcd to the duration of the related video ?

rbouqueau: Not a use-case we have already encountered (I could write more tests to explicit this). But if we concat components (audio, video, tmcd) separately then it will make a difference e.g.:
a/v/tmcd duration 1
+ a duration 1 => a(2)/v(1)/tmcd(2)
+ v/tmcd duration 1 => a(2)/v(2)/tmcd(3) where tmcd is then longer by one...

What do you think?

jeanlf: no it shouldn't be longer since you check for the video ref. Actually this check should prevent from extending all tmcd if you start the iteration from 0 no?

rbouqueau: change pushed accordingly

jeanlf: looks good to me, you can merge (and add a test in testsuite pretty please)

PR #1544, opened 2020-07-17
dash: fix multiple packets with the same cts/dts

mspieth:
fixes dash stream generation for live streams

test command line is for live stream is

`gpac -graph -stats "httpin:FID=i1:src=$1" @ \
	"dashin:FID=d1:algo=none:start_with=max_q:speedadapt=yes:use_bmin=true:lowlat=no" \
	"reframer:FID=rf:SID=d1:rt=on:speed=1.0" \
	"ffmx:FID=o1:SID=rf:dst=out.ts"
`

alternate with udp out ffmx
`
"ffmx:FID=o1:SID=rf:ffmt=mpegts::dst=udp://10.10.1.1:4444::pkt_size=1316:buffer_size=65535"
`

Tested both with vlc

NOTE: It still fails to generate file in realtime with m2tsmx and fout but that is another problem not addressed here. This is why ffmx was used. reframer is optional but helps with udp output stream.

### DIFF
--- a/src/filters/dmx_dash.c
+++ b/src/filters/dmx_dash.c
@@ -88,6 +88,9 @@ typedef struct
 	Bool prev_is_init_segment;
 	u32 timescale;
 	s64 pto;
+	u64 first_cts;
+	u64 first_dts;
+	u64 repeat_ts_offset;
 	s64 max_cts_in_period;
 	bin128 key_IV;
 
@@ -149,7 +152,9 @@ void dashdmx_forward_packet(GF_DASHDmxCtx *ctx, GF_FilterPacket *in_pck, GF_Filt
 			} else {
 				group->max_cts_in_period = 0;
 			}
-
+			group->repeat_ts_offset = 0;
+			group->first_cts = cts;
+			group->first_dts = dts;
 			start = (u64) (scale * gf_dash_get_period_start(ctx->dash));
 			group->pto -= start;
 		}
@@ -162,6 +167,9 @@ void dashdmx_forward_packet(GF_DASHDmxCtx *ctx, GF_FilterPacket *in_pck, GF_Filt
 
 		//remap timestamps to our timeline
 		if (dts != GF_FILTER_NO_TS) {
+			if (dts == group->first_dts) {
+				dts += group->repeat_ts_offset;
+			}
 			if ((s64) dts >= group->pto)
 				dts -= group->pto;
 			else {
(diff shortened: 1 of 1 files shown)


jeanlf:
Thanks for this interesting PR. Couple of comments:
- DTS/CTS should never be the same for a given stream in the same period, any clue what is causing this at the packager side ?
- You need to add a warning whenever same DTS is detected
- Do you know what causes the m2tsmx to fail ?

mspieth:
1. The dash stream being received supplies packets in 3 second chunks. Packets within the chunk all have the same cts and dts. I instrumented the dash_dmx filter to find this out. 3 streams of course A+T+V all with different timescales.
2. This happens all the time as per point 1.
3, Not as yet. It seems to buffer up lots. May be an av sync/interleave issue but I'm not knowledgable enough yet to say.
The patch allows me to view the stream.

There is 1 new issue at about 13 seconds the video corrupts and after that there are no more corruptions. I believe this to be a hard/abrupt AV resync. No audio blips can be heard when this happens, probably because V precedes A when it gets to the mux.
I think this has to do with a full 9 seconds of video gets fed to the mux before any audio and eventually it complains and fixes itself. Ive only tried a 2 min capture so far. Will get more data over the weekend.
I have not tried it much with a static stream post patch. I was testing with big buck bunny to get the gpac filter config right and now I understand that much more.
The double :: trick to escape colons (in urls, incl udp) is not documented but very useful. Trawled through the code for that one.

jeanlf:
> The dash stream being received supplies packets in 3 second chunks. Packets within the chunk all have the same cts and dts.

This is crazy! Could you upload one or more chunks and the init segment somewhere ?

> Not as yet. It seems to buffer up lots. May be an av sync/interleave issue but I'm not knowledgable enough yet to say.
The patch allows me to view the stream.

I think the m2tsmx issue come from the fact that DTS are broken

> The double :: trick to escape colons (in urls, incl udp) is not documented but very useful. Trawled through the code for that one.

It's hidden [here](https://github.com/gpac/gpac/wiki/filters_general#specifying-encoders-and-decoders), but you're right this is hard to find. I will add that to the filter declaration section.

mspieth:
> This is crazy! Could you upload one or more chunks and the init segment somewhere ?
Hopefully this is good enough first one is init, 2nd one is an aac data segment. cat them together and convert to mp4 to play (ffmpeg is works for this)

https://paste.c-net.org/BarnettDuran
https://paste.c-net.org/GillmanToolbox

This is from a commercial streaming portal and a live channel. I do not know how they generate it, but shakaplayer is fine with it.

jeanlf:
hmm links are just text files, did I miss something?

mspieth:
Sorry, try again

https://paste.c-net.org/BiopsyLombardo
https://paste.c-net.org/EvolvedLinen

jeanlf:
The segment is properly formatted with 1024 sample duration (48k timescale), could you get us more data reproducing what you observed ?
You can redirect the dasher to the inspect filter to check this:

`gpac -i source.mpd inspect:deep`

mspieth:
Solved a different way.

PR #1551, opened 2020-07-24
fix build with no zlib

rbouqueau:


jeanlf:
fixing in a different way to be more consistent with gf_gz_compress_payload()

Issue #1606, opened 2020-10-07
mp4box unable to set individual track/stream label metadata in mp4/m4v files


eharris:
I'm trying to use mp4box to set track/stream metadata labels, and it doesn't seem to work.  No error is produced, but the UDTA atom isn't getting set either.  The documentation for `-set-meta` seems like this should work.  Specifically, I'm trying to set the UDTA `name` atom, which is displayed by vlc and other players to identify/differentiate multiple streams.

Example:
```
> mp4box -set-meta 'name;3=Commentary' output.m4v
Setting up iTunes/iPod file...
Saving output.m4v: 0.500 secs Interleaving
> mp4box -info output.m4v
...
Track # 3 Info - TrackID 3 - TimeScale 48000
Media Duration 01:47:56.394 - Indicated Duration 01:47:56.394
Track has 1 edit lists: track duration is 01:47:55.373
Track is disabled
Media Info: Language "English (eng)" - Type "soun:mp4a" - 303581 samples
MPEG-4 Config: Audio Stream - ObjectTypeIndication 0x40
MPEG-4 Audio AAC LC (AOT=2 implicit) - 2 Channel(s) - SampleRate 48000
Synchronized on stream 1
        RFC6381 Codec Parameters: mp4a.40.2
Alternate Group ID 1
        All samples are sync
        Max sample duration: 1024 / 48000
```

Yet this is the info output of a different file produced by Handbrake that has the stream name set correctly (note the UDTA line that was missing in the mp4box produced file):
```
Track # 3 Info - TrackID 3 - TimeScale 48000
Media Duration 01:47:56.394 - Indicated Duration 01:47:56.394
Track has 1 edit lists: track duration is 01:47:56.373
Track is disabled
Media Info: Language "English (eng)" - Type "soun:mp4a" - 303581 samples
1 UDTA types: name (1) 
MPEG-4 Config: Audio Stream - ObjectTypeIndication 0x40
MPEG-4 Audio AAC LC (AOT=2 implicit) - 2 Channel(s) - SampleRate 48000
Synchronized on stream 1
        RFC6381 Codec Parameters: mp4a.40.2
Alternate Group ID 1
        All samples are sync
        Max sample duration: 1024 / 48000
```

Version: 1.0.1-rev0-gd8538e8ae-HEAD (compiled from latest release source)
OS: Debian buster amd64

Reference info:
https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html

[...]

PR #1619, opened 2020-10-11
moved documentation of -udta option to meta section

eharris:
The documentation for the `-udta` option was under the "File Splitting and Concatenation" options, when it seems more appropriate for this to be under the Meta Options section, so moved it there.

### DIFF
--- a/share/doc/man/mp4box.1
+++ b/share/doc/man/mp4box.1
@@ -362,27 +362,6 @@ keep UTC timing in the file after edit
 
 .br
 .TP
-.B \-udta  tkID:[OPTS] (string)
-.br
-set udta for given track or movie if tkID is 0. OPTS is a colon separated list of:
-.br
-* type=CODE: 4CC code of the UDTA (not needed for box= option)
-.br
-* box=FILE: location of the udta data, formatted as serialized boxes
-.br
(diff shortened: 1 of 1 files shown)


jeanlf:
thanks but there is a reason for this classification: the 'meta' help is about the 'meta' box tools in ISOBMFF (mpeg21, HEIF), not about tagging (udta, itags, copyright...). I'll rework the general help intro instead

eharris:
I was mainly trying to get the documentation for the `-udta` option into a location that would make it more likely to get found.  I did quite a lot of searching of both google and the mp4box docs trying to find out how to set the udta name metadata, so I was looking in the pages about meta data (like https://github.com/gpac/gpac/wiki/mp4box-meta-opts ), not about file splitting.  

I was actually going to try updating the wiki page(s) to try to add info on `-udta` as well as examples like you gave in https://github.com/gpac/gpac/issues/1606#issuecomment-706069929 , but since you have indicated that the organization I was thinking was appropriate is not, I will await your documentation changes to see how it should be organized before attempting further changes.

jeanlf:
OK - I think we can indeed re-organize differently the general help, which mixes a lot of different concepts. Any user feedback on that is more than welcome :)

PR #1620, opened 2020-10-12
Updated documentation links

eharris:


jeanlf:
now fixed thanks for the report

PR #1658, opened 2020-12-14
Buildbot hls intra only

rbouqueau:


src/filters/dasher.c
```diff
@@ -1869,6 +1871,9 @@ static Bool dasher_same_adaptation_set(GF_DasherCtx *ctx, GF_DashStream *ds, GF_
 	//not the same roles
 	if (!dasher_same_roles(ds, ds_test)) return GF_FALSE;
 
+	//intra-only trick mode belongs to a separate AS
+	if (ds->hls_intra_only != ds_test->stream_type) return GF_FALSE;
```
jeanlf, line 1875: shouldn't that be 
` if (ds->hls_intra_only != ds_test->hls_intra_only) return GF_FALSE;`

src/filter_core/filter_props.c
```diff
@@ -1359,6 +1359,7 @@ GF_BuiltInProperty GF_BuiltInProps [] =
 	{ GF_PROP_PID_CLAMP_DUR, "ClampDur", "Max media duration to process from pid in DASH mode", GF_PROP_DOUBLE, GF_PROP_FLAG_GSF_REM},
 	{ GF_PROP_PID_HLS_PLAYLIST, "HLSPL", "Name of the HLS variant playlist for this media", GF_PROP_STRING, GF_PROP_FLAG_GSF_REM},
 	{ GF_PROP_PID_HLS_GROUPID, "HLSGroup", "Name of HLS Group of a stream", GF_PROP_STRING, GF_PROP_FLAG_GSF_REM},
+	{ GF_PROP_PID_INTRA_ONLY, "IntraOnly", "Indicates the media contains intra frames only (SAP=1)", GF_PROP_BOOL, GF_PROP_FLAG_GSF_REM},
```
jeanlf, line 1362: I think 'GF_PROP_PID_HAS_SYNC' is already signaling this: if present and FALSE, all samples are sync, if present and TRUE, some are sync some are not, if absent, unknown state.
Could you reuse this property instead of defining a new one ?

jeanlf (changes requested):

rbouqueau:
@jeanlf comments addressed

src/filters/reframer.c
```diff
@@ -241,7 +241,7 @@ GF_Err reframer_configure_pid(GF_Filter *filter, GF_FilterPid *pid, Bool is_remo
 	for (i=0; i<ctx->saps.nb_items; i++) {
 		switch (ctx->saps.vals[i]) {
 		case 1: ctx->filter_sap1 = GF_TRUE;
-			gf_filter_pid_set_property(st->opid, GF_PROP_PID_INTRA_ONLY, &PROP_BOOL(GF_TRUE));
+			gf_filter_pid_set_property(st->opid, GF_PROP_PID_HAS_SYNC, &PROP_BOOL(GF_FALSE)); //false: all samples are sync
 			break;
 		case 2: ctx->filter_sap2 = GF_TRUE; break;
```
jeanlf, line 246: could you also add that for SAP2 (in case we find such a packet one of these days) - it's technically still closed gop.
The rest looks good to me thanks !

rbouqueau: addressed

PR #1668, opened 2020-12-31
filters/reframe_qcp.c: fix potential crash

XidongWang-coder:
In function qcpdmx_configure_pid() in file filters/reframe_qcp.c, ctx->opid should be checked before passed to gf_filter_pid_remove
in case of potential NULL pointer dereference.


jeanlf:
thanks, fixed as part of a more general cleanup on pid removal

Issue #1708, opened 2021-03-15
[FEATURE REQUEST] Migration to Gradle/POD


Duna:
I just want to thank you for your open source project, it is awesome!

One more thing I want to add: Migrate the Android/iOS integration to Gradle/PODs. It is very **hard** to get a successful .so files and create wrappers. FFMPEG has one project with gradle integration for Android and iOS, You should do the same.

Gradle file:
```
dependencies {
    implementation 'com.arthenica:mobile-ffmpeg-full-gpl:4.4.LTS'
}
```
Same with iOS project by including PODs

aureliendavid:
I don't think uploading to any sort of central repository is in the books for now since the android versions are very much works in progress. 

We do upload binary versions of osmo4 for android at https://gpac.wp.imt.fr/downloads/gpac-nightly-builds/ (these apks will now include libmp4box.so since it's built at the same time). Building mp4box for android from there shouldn't be too. We might also distribute binary version of mp4box for android as apks in the future but it probably will be on gpac.io and not on gradle or something similar. 

Although if we do find development resources for android and the projects acquire more maturity, I agree that it would be a good distribution option then.

Issue #1693, opened 2021-02-28
[Need Info] Android MP4box command line seems to do nothing


Duna:
We have successfully compiled the gpac on ARM/ARM_V7/x86 using Android NDK.
We created a new Android Studio 4.1 project compatible with the newest Android tools, by following the **mp4box_android** sample project.
The new Android project can be downloaded from [here](https://drive.google.com/file/d/1R0iV5oK9KSArBYV0CsFCU1CYwXQ9RBrp/view?usp=sharing).

We granted the **write_external_storage** runtime permission and set `android:requestLegacyExternalStorage="true"` in the AndroidManifest.xml since we are using `compileSdkVersion 30`

We added the sample video onto device file within the Android Download folder [legend.mp4](https://drive.google.com/file/d/122-BJc9lJOHssyM_hLFHObi8XrZlHgoT/view?usp=sharing)

Within the Android app we have tried to load successfully all the .so libs compiled
Within the Android project we have tried to execute the following command for mp4box: "`MP4Box -frag 240000 /storage/emulated/0/Download/legend.mp4 -out /storage/emulated/0/Download/legend_frag.mp4`"
We get the following console output:
```
2021-02-28 13:26:29.592 9274-9349/com.enst.mp4box V/WRAPPER_JNI: mp4terminal::start
2021-02-28 13:26:29.593 9274-9349/com.enst.mp4box V/WRAPPER_JNI: mp4terminal::command get back ok
2021-02-28 13:26:29.593 9274-9349/com.enst.mp4box V/WRAPPER_JNI: Convert command line done
2021-02-28 13:26:29.622 9274-9349/com.enst.mp4box V/WRAPPER_JNI: mp4terminal::end
```
**Current** result:
 - no output video file 
 
**Expected** result:
 - /storage/emulated/0/Download/legend_frag.mp4 should be created

We have tried different command lines but with no good result. 
One thing I have to note here, when doing -frag or some other command lines, the terminal should pop the message `WRAPPER_JNI: mp4terminal::end` only after ~10seconds since the video processing takes a while, depending of how long the input video is. Currently `mp4terminal::end` comes right after `mp4terminal::start` no matter which command line is executed

Moreover we put the .so files in the osmo4 project and is working fine

Duna:
The question are: 
- Am I using this command in the right way? 
- This way should I provide the path to the input files?

[...]

Issue #1691, opened 2021-02-25
Text relocation error for Android+x86 architecture


Duna:
I am trying to compile the .so libs for an android project, unsuccessful because Text Relocation issue on newer android sdk
It seems this it needs to be fixed in x86 libs

rbouqueau:
It is difficult to answer without any detail, could you provide more?

Usually these errors are solved by using -fPIC.

Duna:
> It is difficult to answer without any detail, could you provide more?
> 
> Usually these errors are solved by using -fPIC.

No, it didn't work. disable-asm  worked, but I am wondering if the code still working

rbouqueau:
> disable-asm worked

GPAC has no such option. Would you please provide us with the log to understand more about the error?

ATiwari1987:
2021-02-25 14:43:27.414 20619-20619/com.enst.mp4box I/System.out: Loading library avcodec...
2021-02-25 14:43:27.415 20619-20619/com.enst.mp4box I/System.out: dlopen failed: "/data/app/~~ZmEO3tVu7m6NJGkYR6wdNQ==/com.enst.mp4box-E4Luw3FTCH87NiWcgTfBdg==/lib/x86/libavcodec.so" is 64-bit instead of 32-bit

ATiwari1987:
2021-02-25 13:43:45.907 26423-26423/com.enst.mp4box I/System.out: Loading library avcodec...
2021-02-25 13:43:45.907 26423-26423/com.enst.mp4box E/linker: "/data/app/~~BXk9ztSoipkJMdbtaI3cfw==/com.enst.mp4box-SvfB-NjbdSVdQ1kn6aJM8A==/lib/x86/libavcodec.so" has text relocations (https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#Text-Relocations-Enforced-for-API-level-23)

Getting this error when loading avcodec library in gpac project. However this  issue is only on x86 and not arm architecture, and gets resolvedby using --disable-asm flag (not with -fPic though)

rbouqueau:
> I am trying to compile the .so libs for an android project, unsuccessful because Text Relocation issue on newer android sdk

Compile error.

> 2021-02-25 13:43:45.907 26423-26423/com.enst.mp4box I/System.out: Loading library avcodec...

Runtime error. This is not a compile error.

Could you please provide way more context about what you did? Imagine that I'd like to help you and reproduce your issue: what do I need to do?

[...]

Issue #1690, opened 2021-02-25
Failed to get libmp4box.so compiled for Android


Duna:
I have followed the tutorial from:
https://github.com/gpac/gpac/wiki/GPAC-Build-Guide-for-Android
I have successfully built for NDK:
`<GPAC_ROOT_DIR>/deps_android/build/android$ ./gpac_build_all_extra_libs /opt/android-ndk` having the following .so libs compiled:
```
duna@duna-VirtualBox:~/gpac_public/extra_lib/lib/android/armeabi$ ls
libavcodec.so   libavformat.so  libfaad.so   	libjs_osmo.so   libpng.so         	libswscale.so
libavdevice.so  libavutil.so	libft2.so    	libmad.so   	libstlport_shared.so  libz.so
libavfilter.so  libeditline.so  libjpegdroid.so  libopenjpeg.so  libswresample.so
```

The main issue is I do not manage to get libmp4box.so compiled file on Android. 
I executed the command line:
`<GPAC_ROOT_DIR>/gpac_public/build/android/jni$ ./gpac_build_android -ndk=/opt/android-ndk -sdk=/opt/android-sdk -jdk=/usr/lib/jvm/java-8-openjdk-amd64 -force_rebuild`

Built the library for armeabi but the output is:
```
duna@duna-VirtualBox:~/gpac_public/applications/mp4box_android/libs/armeabi$ ls
libeditline.so  libgpac.so   	libjs_osmo.so   libpng.so
libft2.so   	libjpegdroid.so  libopenjpeg.so  libz.so
```

Duna:
How can I get the libmp4box.so compiled for Android?

aureliendavid:
libmp4box doesn't exist, the main library is libgpac

mp4box and osmo4 are programs that both use libgpac to do different things

PR #1709, opened 2021-03-16
MP4Box for Android

aureliendavid:
[ references to issues: #1687 #1688 #1689 #1690 #1691 #1693 #1708 ]

Hi all, 

I've updated our build scripts to be able to build mp4box for android. 

### What's done: 

- fix ndk build of `libmp4box.so` for arm/x86 by the `build/android/jni/gpac_build_android` script
- command line gradle build of mp4box apk in the same script
- update mp4box android project to be buildable with android studio

### Limitations

- I've kept our slightly old versions of the android environment; first because I wanted to change as little as possible and because I think it's easier to update the projects to newer environments than to lose backward compatibility. 

The versions used here are: ndk r13b, sdk api v25, android studio 2.2.3

- The MP4Box android port currently consists of a simple gui with a textbox for the command line arguments and stderr/stdout redirected to text files in `/mnt/sdcard` : 

mp4boxandroid

It's a good proof of concept but is pretty much a usability nightmare for serious use. 

If there is interest in the project it would be nice to either convert it to command line tool or to update the gui to at least have the output displayed in the app. 

That being said, now that the project should build easily it will be easier to improve it in the future (contributions welcomed). 

- I haven't added automatic public distribution of the mp4box apks like we do with osmo4. It would be doable if there is interest in it. However `libmp4box.so` will be included in the libs of the Osmo4 apk so people could just take these libs and build the java project with android studio. 

That's where we are right now. I will merge this PR soon if there are no objections to it.


Fatima212:
I have successfully compiled MP4Box_android. But the problem is that I can only run the command line one time after the app was opened. If I run the command one more time, I will got this message in _/mnt/sdcard/stderrout.txt_ : "Error - 2 input names specified, please check usage"

aureliendavid:
@Fatima212 you're right, it doesn't work properly on the second execution

this version was just a POC of a jni wrapper around mp4box, the problem is that, as a command line tool, the mp4box main() is not made to be called more than once (because there are a bunch of global variables that are not reset between calls)

I think the better solution would be to build a proper executable MP4Box command line for android, and maybe build a java gui that just calls it; but running and bundling binaries is a bit of a pain on android

what I did in the above commit is a hack in which the app restarts itself after each execution of mp4box, I also added a field to display the log of the previous execution: 

Screenshot_2022-03-01-17-17-35-520_com gpac mp4box

it's a bit clunky but it kinda works

Fatima212:
@aureliendavid Thank you for your help! 
After reading some .c files, I figure out this error occurs because `inName` variable  is not reset. https://github.com/gpac/gpac/blob/7b64a5c3b51abd225d6e869e8e8a8057d18a07cb/applications/mp4box/main.c#L245
I have set `inName = NULL` in first line of `mp4boxMain` function so that `inName ` is reset every time mp4boxMain is called. 
https://github.com/gpac/gpac/blob/7b64a5c3b51abd225d6e869e8e8a8057d18a07cb/applications/mp4box/main.c#L5671
Rebuilding mp4box libs, my mp4box_android application can run without restart. 
Maybe it would help.

aureliendavid:
yes in this particular it was inName that was not reset and caused the "2 input files" error

but in the general case there might be other stuff that are not reinitialized between runs, so I figured reloading the whole thing safer

PR #1710, opened 2021-03-17
AV1: Use sequence header resolution if frame override is not set.

tdaede:
Corrects frame sizes for scalable images, among other applications.


jeanlf:
thanks !

PR #1712, opened 2021-03-20
fix AV1 timing info parser

cconcolato:


rbouqueau:
Thanks for submitting! LGTM

PR #1717, opened 2021-03-25
AVIF: Add a1lx box support.

tdaede:
This automatically adds an a1lx box if the input image has
multiple spatial layers.


include/gpac/isomedia.h
```diff
@@ -3204,6 +3204,14 @@ GF_VVCConfig *gf_isom_vvc_config_get(GF_ISOFile *isom_file, u32 trackNumber, u32
 */
 GF_AV1Config *gf_isom_av1_config_get(GF_ISOFile *isom_file, u32 trackNumber, u32 sampleDescriptionIndex);
 
+/*! gets AV1 scalable layer byte offsets for a1lx box
+\param isom_file the target ISO file
+\param trackNumber the target track
+\param layer_size returned 4 layer sizes
+\return none
+*/
+void gf_isom_av1_layer_size_get(GF_ISOFile *file, GF_ISOTrackID trackID, u32 layer_size[4]);
```
jeanlf, line 3213: could you make this function return GF_Err, BAD_PARAM if wrong params or not AV1 track, ISOM_INVALID_FILE if no av1 config ?

src/isomedia/box_funcs.c
```diff
@@ -1307,6 +1310,9 @@ static struct box_registry_entry {
 	BOX_DEFINE_S(GF_ISOM_BOX_TYPE_CLLI, clli, "mp4v jpeg avc1 avc2 avc3 avc4 svc1 svc2 hvc1 hev1 hvc2 hev2 lhv1 lhe1 vvc1 vvi1 encv resv", "miaf"),
 	BOX_DEFINE_S(GF_ISOM_BOX_TYPE_MDCV, mdcv, "mp4v jpeg avc1 avc2 avc3 avc4 svc1 svc2 hvc1 hev1 hvc2 hev2 lhv1 lhe1 vvc1 vvi1 encv resv", "miaf"),
 
+	//AVIF
+	FBOX_DEFINE_S(GF_ISOM_BOX_TYPE_A1LX, a1lx, "ipco", 0, "iff"),
```
jeanlf, line 1314: could you change "iff" to "avif" ?

jeanlf (changes requested):
Great, many thanks, looks almost perfect!

Do you have a public AV1 multi layer source file we could use for the test suite ?

tdaede:
While implementing support for u16 layer sizes I realized that the current spec is a bit inefficient: https://github.com/AOMediaCodec/av1-avif/issues/140

I'd like to resolve this before landing.

tdaede:
OK, so I've now updated the a1lx box to reflect the new spec changes. Specifically:
* It's now a simple box rather than a full box
* It is only 3 entries rather than 4
* Last entry is always 0
* u8 length field replaces flag

PR #1730, opened 2021-04-06
dasher: handler errors instead of asserting when the number of stream at period switch is beyond expectation

rbouqueau:


rbouqueau:
Can we merge?

PR #1744, opened 2021-04-14
fix parsing of layered AV1 streams with varying tile config

cconcolato:
Current MP4Box cannot parse multi-layered streams where the resolution and tiling changes. This fixes the bug.


rbouqueau:
LGTM. Do we have a plan to add a regression test? I can help strip the file if it makes sense.

PR #1748, opened 2021-04-20
AVIF: Add basic support for a1op.

tdaede:
This will not write the actual operating point selector unless
it is manually populated into the image properties struct.


jeanlf:
This seems to include al1x, can we close #1717 ?

jeanlf:
I merged #1717 and did a few cleanups, plus added tests in [testsuite](https://github.com/gpac/testsuite/blob/master/scripts/iff-misc.sh), you can rebase.

jeanlf:
unless I'm mistaking, this only has the code to read/write the box but no option to set the a1op value right ?

tdaede:
I had forgotten to push the latest version of the code. There is a CLI parameter for it.

One thing that is not plumbed is passing the operating point to the AV1 parsing code - so if the operating points should result in other properties having different values, that is not reflected. I can plumb it through into all the av1 parsing code, but wanted to get feedback on whether you liked the current way of setting the option first.

jeanlf:
seems OK, but we should check that the specified value is less than operating_points_cnt_minus_1 (stored in state->operating_points_count) no?

tdaede:
Updated to have a check for exceeding operating_points_count.

jeanlf:
I think that's good. I can merge now if you want, and we'll add the update of item properties depending on the chosen OP later. Unless you have something ready ?

tdaede:
Yeah, I think it's useful as is, I'd like to merge it now.

jeanlf:
merged, thanks !

PR #1750, opened 2021-04-20
Verify JFIF markers before reading

gorinje:
On some JPEG files, the 0xFF + 0xC0/C1 markers could be contained before the actual start of frame. This may lead to inconsistency between the size extracted from rfimg filter and the one from imgdec (in my case it crashes the display).

This is a very small fix taken from this description :
http://vip.sugovica.hu/Sardi/kepnezo/JPEG%20File%20Layout%20and%20Format.htm

where Length should be equal "to 8 + components*3 value" in case it could be usefull to you.

Enclosed is a jpeg file to reproduce this error.
bird


jeanlf:
Thanks, I'll add the file to the test suite as well

PR #1760, opened 2021-04-26
fix cmaf compliance: mehd should be absent when duration is unknown

rbouqueau:


src/filters/mux_isom.c
```diff
@@ -4455,6 +4455,14 @@ static GF_Err mp4_mux_initialize_movie(GF_MP4MuxCtx *ctx)
 		gf_isom_set_movie_duration(ctx->file, mdur);
 	}
 
+	if (ctx->cmaf && max_dur.num == 0) {
+		//CMAF 7.3.2.1.c.6) "The MovieExtendsBox may contain a MovieExtendsHeaderBox,
+		//as defined in ISO/IEC 14496-12, and if so, shall provide the overall duration
+		//of the CMAF track. If the duration is unknown, this box shall be omitted."
+		gf_isom_box_del_parent(&ctx->file->moov->mvex->child_boxes, (GF_Box*)ctx->file->moov->mvex->mehd);
```
jeanlf, line 4462: Accessing movie internals should not be done outside libisomedia, could you add a wrapper for that, or add a boolean to gf_isom_set_movie_duration to force mehd removal ?

jeanlf (changes requested):

src/filters/mux_isom.c
rbouqueau: Done. My issue was that ```gf_isom_set_movie_duration()``` is not called.

jeanlf:
merged from first commit, didn't want to modify gf_isom_finalize_for_fragment

PR #1801, opened 2021-05-25
fix NULL pointer dereference in gf_media_export_saf

hifoolno:
There is a similar fix in https://github.com/gpac/gpac/commit/87afe070cd6866df7fe80f11b26ef75161de85e0


jeanlf:
This will create a memory leak by not destroying esd.
Patched, thanks for the report!

PR #1802, opened 2021-05-25
Improve the desktop bits

pinotree:
- install the application icon in the XDG hicolor icon theme, rather than the legacy pixmaps location
- make `gpac.desktop` file static, as it does not change anymore
- drop an extra key in `gpac.desktop`


jeanlf:
Thanks, merged !

PR #1867, opened 2021-08-01
fix parse date

gmaen:
fixed:
when parse '2021-07-31T12:50:20.000+08:00',
1)
(sscanf(val, "%d-%d-%dT%d:%d:%gZ", &year, &month, &day, &h, &m, &secs) == 6) is TRUE,
then oh = om = 0, 
should into 
(sscanf(val, "%d-%d-%dT%d:%d:%g+%d:%d", &year, &month, &day, &h, &m, &secs, &oh, &om) == 8)
2)
utc of '2021-07-31T12:50:20.000+08:00' is 2021-07-31 04:50:20,
so current_time should minus diff.

### DIFF
--- a/src/utils/os_divers.c
+++ b/src/utils/os_divers.c
@@ -2553,12 +2553,12 @@ u64 gf_net_parse_date(const char *val)
 	oh = om = 0;
 	secs = 0;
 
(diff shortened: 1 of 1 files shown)


jeanlf:
thanks,merged

Issue #1913, opened 2021-09-15
Wrong CTS calculation for VVC bitstreams 


K-os:
Thanks for reporting your issue. Please make sure these boxes are checked before submitting your issue - thank you!

- [x] I looked for a similar issue and couldn't find any.
- [x] I tried with the latest version of GPAC. Installers available at http://gpac.io/downloads/gpac-nightly-builds/
- [x] I give enough information for contributors to reproduce my issue (meaningful title, github labels, platform and compiler, command-line ...). I can share files anonymously with this dropbox: https://www.mediafire.com/filedrop/filedrop_hosted.php?drop=eec9e058a9486fe4e99c33021481d9e1826ca9dbc242a6cfaab0fe95da5e5d95

Detailed guidelines: http://gpac.io/2013/07/16/how-to-file-a-bug-properly/

----

Hi,

I am having trouble packaging a VVC bitstream using the current master branch. The CTS timestamp for the encoded frames jumps, when a CRA or IDR frame with leading pictures is encountered. The problem lies somewhere in the parsing of the bitstream.

For example, given a bitstream  with GOP size 16 and an intra period of 32 I get the following output, when inspecting the raw bitstream (`bin/gcc/gpac -i BasketballPass_QP37.266:fps=50 inspect:deep`):
```
PID 1 name BasketballPass_QP37.266 Configure - properties:
	SourcePath: BasketballPass_QP37.266
	URL: BasketballPass_QP37.266
	Extension: 266
	Cached: true
	DownloadSize: 22062
	StreamType: Visual
	ID: 1
	Width: 416
	Height: 240
	FPS: 50
	Timescale: 50
	CodecID: VVC Video
	DecoderConfig: 176 bytes (CRC32 0xDF88D241)
	PlaybackMode: forward
PID 1 PCK 1 - full frame dts 0 cts 0 dur 1 sap 1 ilace 0 corr 0 seek 0 bo N/A roll 0 crypt 0 vers 0 size 6212 lp 0 depo 0 depf 0 red 0 CRC32 0x9B9BE30C
PID 1 PCK 2 - full frame dts 1 cts 16 dur 1 sap 0 ilace 0 corr 0 seek 0 bo N/A roll 0 crypt 0 vers 0 size 1783 lp 0 depo 0 depf 0 red 0 CRC32 0x7F0D0484
[...]

PR #1914, opened 2021-09-16
fix POC reset for IDR with leading pics (#1913)

K-os:
I needed to set `slice_is_idr = 0` for IDR_W_LP slices, also. I am not sure if that is correct or produces any other problems, but it fixed the CTS calculation for me.


K-os:
Still some remaining problems with the POC calculation. See #1913

PR #1927, opened 2021-10-03
Export filter register manipulation symbols

RodolpheFouquet:
Hello!

In order to enable the usage of static modules, we need to export those two functions to add and remove filters from the registry.


rbouqueau:
If ok then we also need to export them in export.cpp (for msvc).

PR #1930, opened 2021-10-07
detect VVC bitstream when it starts with an AUD

K-os:
VVC bitstreams starting with an access unit delimiter were not correctly detected as such.


jeanlf:
Looks good to me, but could you add a comment (eg "for files with no vps ....") ?

jeanlf:
Thanks !

PR #1981, opened 2021-12-13
AV1 parsing: only parse qm_v if sperarate_uv_delta_q is true

aseixasd:
I think there is a small bug in parsing AV1 streams with q matrices - see section 5.9.12 of the spec


jeanlf:
Indeed, thanks for the patch !

PR #2085, opened 2022-01-31
Fix fragmented mp4 with composition offset having too short duration

inrustwetrust:
When remuxing a fragmented MP4 that uses B-frames with MP4Box, the duration of the resulting file becomes too short. The amount of missing frames at the end is the same as the number of B-frames.

Reproduction steps:

First, generate a fragmented MP4 with ffmpeg (the version I used was 4.4.1 installed from Brew on OSX). The duration is 1 second/25 frames.

```
ffmpeg -y -f lavfi -i "testsrc=duration=1s" -pix_fmt yuv420p -c:v libx264 -profile high -bf 2 -movflags frag_keyframe+empty_moov+delay_moov+skip_trailer fragmented.mp4
```

Then, remux it with MP4Box:
``` 
MP4Box -add fragmented.mp4 -new remuxed.mp4
```

The resulting `remuxed.mp4` file now has an edit list entry with segment_duration set to 0.92 seconds, i.e. it’s missing two 40ms-frames.

The issue appears to be in the calculation of the `traf` duration in the `MergeTrack()` function. It subtracts the media_time of the source file edit list entry from its duration, but it does not compensate for the fact that the composition start time of the video track may not be 0 when B-frames are used. In this particular example, the source video track starts at composition time 1024 (each frame has duration 512 time base units). Since the source edit list entry media_time is 1024 as well, it means that it effectively points to the very beginning of the video track.

To solve this, I believe that the composition start time should be subtracted from the media_time, and that result is what should be subtracted from the duration. Continuing on the example, this means that the edit list entry media_time and video track start times precisely cancel each other out, and nothing is subtracted from the duration.

The supplied patch implements this logic, by repurposing the already existing `gf_moof_get_earliest_cts()` function from `isomedia/movie_fragments.c` in order to find the correct composition start offset. I don’t have any previous experience with this code base, so I’m not sure if this is a good way to solve things architecturally. I’d be happy to take suggestions for alternatives if not.

### DIFF
--- a/include/gpac/internal/isomedia_dev.h
+++ b/include/gpac/internal/isomedia_dev.h
@@ -4644,6 +4644,8 @@ GF_Box *gf_isom_clone_config_box(GF_Box *box);
 GF_Err gf_isom_box_dump(void *ptr, FILE * trace);
 GF_Err gf_isom_box_array_dump(GF_List *list, FILE * trace);
 
+u64 gf_moof_get_earliest_cts(GF_MovieFragmentBox *moof, GF_ISOTrackID refTrackID);
+
 void gf_isom_registry_disable(u32 boxCode, Bool disable);
 
 /*Apple extensions*/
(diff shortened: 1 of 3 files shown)


jeanlf:
Thanks for the detailed report. I fixed that differently to also check for cases where tfdt extends previous sample duration, this should now work properly.

PR #2252, opened 2022-09-05
fix 64 bit chunk offsets when appending

ahaensler:
I noticed that when [merging tracks](https://github.com/gpac/gpac/blob/master/src/isomedia/track.c#L468), gpac gave up after 2^32 bytes.

Looks to me like a small bug in `stbl_AppendChunk`. One line got lost in the 64-bit codepath. The PR adds that line and now gpac works as expected when merging many tracks.


jeanlf:
Indeed, thanks for the patch !

Issue #1490, opened 2020-05-16
Windows XP  - gpac-0.8.0-rev1-gc1990d5c-master-win32.exe 


qa58:
With 0.8.0 couldn't find any notation of change of supported Windows operating systems, used PETools to peek how 0.8.0 exes were compiled, was for Windows 6 which is Vista - was that intentional or a mistake? I patched mp4box back to Windows 5.1 (XP) and it runs in XP - except I don't know if the DLLs contain Vista+ Windows API calls and will choke.
Personally, I like to encourage use of Windows 7 SDK documentation so code remains compatible with most versions of Windows - Microsoft makes additions into the Windows API that often I find are not really necessary and if used, serve only to break Vista, XP, etc compatibility.

aureliendavid:
I've looked into this a bit. 

I think the short answer is: XP isn't fully supported anymore, but it still might work. 

Longer answer: 

(there's an interesting similar discussion here: https://github.com/sqlitebrowser/sqlitebrowser/issues/893) 

## PE Header

A while ago we went from Visual Studio 2010 to 2015. The new default since VS2013 for the `MinimumRequiredVersion` field is 6. This value in itself is just a flag in the PE header and can easily be changed before or after compilation. 

## XP compatibiliy 

Now for the more complex question of real XP compatibility, there are 3 major obstacles in our recent codebase:

1. We abandoned enforcing C89 standard in favor of C99 (most notably for declarations after statements and designated initializers). This prevents from using VS2010, the v100 toolset and the 7.1SDK (*). 

2. Incompatible 3rd-party dependencies: in particular ffmpeg dropped xp support so everything linked against ffmpeg libraries won't work (e.g.: libavutil depends on bcrypt.dll which doesn't exist for xp: https://patchwork.ffmpeg.org/project/ffmpeg/patch/20180330124452.4800-1-robux4@ycbcr.xyz/ )

3. Use of more recent windows api calls

[_(*) To clarify; that doesn't mean it won't work on windows 7. It will. The windows 7.1 sdk is just the last version to provide full xp support_]

Now to add another layer of complexity: these 3 factors impact our 2 main gpac version (0.8.x legacy branch, and 0.1.x master branch) very differently. 

## legacy branch (0.8.x)

For legacy, as mentioned in the OP, once the PE header is modified, the binaries that we distribute in our installer mostly work. 

The `gm_ffmpeg_in.dll` module won't work (for reason 2.) but `MP4Box` seem pretty much ok. 

[...]

PR #2257, opened 2022-09-14
add support for building mp4box on windows xp

aureliendavid:
this is based on the article at http://wp.xin.at/archives/5810

see also previous discussion about xp support here: https://github.com/gpac/gpac/issues/1490#issuecomment-649519836

this allows building the "mp4box only" profile with xp compatible toolsets v140_xp or v141_xp

(it doesn't allow for a full build on xp, nor does it makes the official installers xp compatible)

tested with v140_xp on VS2015 and x141_xp on VS2019 on a xp sp3 x86 vm

<hr />

the added support for inet_ntop on xp might be overkill, we might just want to undefine GPAC_HAS_IPV6 on xp and not bother with the extra code, but it's debatable

### DIFF
--- a/include/gpac/setup.h
+++ b/include/gpac/setup.h
@@ -169,6 +169,13 @@ typedef unsigned int size_t;
 
 #define snprintf _snprintf
 
+/*	the _USING_V110_SDK71_ macro will be defined when using
+	msvc toolsets like v140_xp, v141_xp, etc.
+*/
+#if (WINVER <= 0x0502 || _USING_V110_SDK71_)
+#define GPAC_BUILD_FOR_WINXP
+#endif
+
 #endif	/*END WIN32 non win-ce*/
 /*end WIN32 config*/
 
--- a/include/gpac/thread.h
+++ b/include/gpac/thread.h
@@ -68,6 +68,55 @@ The thread object allows executing some code independently of the main process o
 #include <windows.h>
 #include <winbase.h>
 
+
+#ifdef GPAC_BUILD_FOR_WINXP
+
+/* Addendum by M. Lackner: Wrap InterlockedExchangeAdd64() around
+   InterlockedCompareExchange64() and name it specifically as an XP
+   function for 32-bit builds only */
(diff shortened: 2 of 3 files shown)


Noitarud:
I will list some ways I thought it could be implemented that would not involve the publisher:
- a converter, it can incorporate missing functions elsewhere in executable or in separate DLL. Could make Ollydbg to import such a file then export - would allow troubleshooting (does not recognise it at present).
- or I thought could use Wine, I suggested a windows port to Wine since I noticed it ran new programs the old windows could not run (due to missing DLLs), I suggested this to them but you could almost feel their shuddering at such an idea (perhaps would upset MS, but MS made a converter to allow old office to open new office files…).

Noitarud:
There are other ways it can be done, if i were to develop another application i would develop it for a Virtual machine, because operating systems are updated and things break, then you need to baybysit it for the life of the application, and it wont work after you leave it and it falls into disuse when the OS is retired. Just too much work. VMs allow easy distribution to other platforms, if there are new bugs you leave it to the afficionados to fix them, not yourself.
Here are a short list of VMs for this or other new apps from your ingenuity:

- x86 pc emulator*
- gameboy
- scumm
- J2ME

* Normally this is used to install an operating system and has aweful performance - but if you build an embodiment like Wine into it or even more lean# then the performance would be radically improved since there are few other processes in parallel.
# first drive contains emulator sw, second drive is the zip of the application, third is application storage image, system settings, registry. When initiated it goes direct into default EXE or a command prompt to launch sister applications (so basically an equipped DOS). I was envisioning one VM per application, run more VMs if want more.

Be aware there is a .js release of MP4BOX but the js system is still under reinvention and i cannot guarantee it would work on XP… https://github.com/gpac/mp4box.js/#build

Noitarud:
There is a .js release of MP4BOX but the js system is still under reinvention and i cannot guarantee it would work on XP… https://github.com/gpac/mp4box.js/#build …still a worthy alternative.

PR #2271, opened 2022-09-29
Vui timing info rewrite

rbouqueau:
To generate a sample with vui timing info (and pic timing sei):

```
gpac -i avgen:dur=2 @ c=avc:r=30/1::x264-params="level=40:nal-hrd=vbr:vbv-bufsize=1000:vbv-maxrate=2000" @ -o avc30.mp4
```

### DIFF
--- a/src/media_tools/av_parsers.c
+++ b/src/media_tools/av_parsers.c
@@ -6359,7 +6359,7 @@ static u8 avc_hevc_get_sar_idx(u32 w, u32 h)
 	return 0xFF;
 }
 
-static void avc_hevc_vvc_rewrite_vui(GF_VUIInfo *vui_info, GF_BitStream *orig, GF_BitStream *mod, Bool is_vvc)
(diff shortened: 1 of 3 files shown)


src/media_tools/av_parsers.c
```diff
@@ -6516,7 +6605,7 @@ static void avc_hevc_vvc_rewrite_vui(GF_VUIInfo *vui_info, GF_BitStream *orig, G
 		}
 	}
 
-	if (is_vvc && vui_chroma_loc_info_present_flag) {
+	if (codec == GF_CODECID_VVC && vui_chroma_loc_info_present_flag) {
```
jeanlf, line 6608: could you put parenthesis around this check ? Apart from that looks good to me

jeanlf (changes requested):

src/media_tools/av_parsers.c
rbouqueau: done

jeanlf:
thanks !

PR #2312, opened 2022-11-11
fix av1 open gop parsing

cconcolato:
Without this fix, when using an Open GoP sequence, the `RefOrderHint` gets corrupted after the "delayed random access point", creating bitstream decoding errors in `skip_mode_params`, and after, including tile group headers.

### DIFF
--- a/src/media_tools/av_parsers.c
+++ b/src/media_tools/av_parsers.c
@@ -3367,7 +3367,7 @@ static void av1_parse_uncompressed_header(GF_BitStream *bs, AV1State *state)
 		if (frame_state->show_existing_frame == GF_TRUE) {
 			frame_state->frame_to_show_map_idx = gf_bs_read_int_log(bs, 3, "frame_to_show_map_idx");
 			frame_state->frame_type = state->RefFrameType[frame_state->frame_to_show_map_idx];
-
+			frame_state->order_hint = state->RefOrderHint[frame_state->frame_to_show_map_idx];
 			if (state->decoder_model_info_present_flag && !state->equal_picture_interval) {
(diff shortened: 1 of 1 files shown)


cconcolato:
cc @agrange

rbouqueau:
This seems ok. Have you checked if other state parameters (```RefValid```, maybe also ```OrderHints```, ```RefUpscaledWidth``` and ```RefFrameHeight```) are also touched by this issue?

rbouqueau:
Questions:
- For the AV1 makers: in AV1, what do we need to preserve from one frame to another?
- In GPAC: should ```order_hint``` belongs to ```frame_state```? Is the reset of the states correct?

agrange:
Hi Romain,

Answers from Jingning (cc'ed) inline


>    - For the AV1 makers: in AV1, what do we need to preserve from one
>    frame to another?
>
> [Jingning] We need to preserve probability models, previous frame's motion
vectors and reference frame indexes on a per block basis, its reference
frames, and frame counter, e.g. order hint.

>
>    - In GPAC: should order_hint belongs to frame_state? Is the reset of
>    the states correct?
>
> [JIngning] Yes.

Hope this helps,
Adrian

rbouqueau:
Ok then the current code has probably a bug and this PR doesn't solve it in a complete way. I see that ```order_hint``` is already [saved under some circulstances](https://github.com/gpac/gpac/blob/05eaac875354682942b70c790bcd62cb5f4cc825/src/media_tools/av_parsers.c#L3202) in ```av1_decode_frame_wrapup()```.

With your sample, where is ```order_hint``` actually reset? This is probably the place where a fix should be applied.

rbouqueau:
What is the status of this PR?

PR #2369, opened 2023-01-03
Adding vui timing info to logs

Lis-TV:
Minor change, extend logging of VUI timing info from SPS


rbouqueau:
LGTM but shouldn't we extend that to HEVC and VVC as well?

Lis-TV:
ok, I'll add it tomorrow.

Lis-TV:
Please integrate AVC printing info only. 
Extension of HEVC/VVC needs much more effort and correct test materials. 
The vui has different structure in AVC/HEVC/VVC. 
I will work on that when I find more time.

PR #2415, opened 2023-03-09
fix using both -dnal and -std on AV1 input

dimitry-unified-streaming:
If you use both -dnal track_id and -std on a AV1 input file, you will always receive an error "Track #1 is not NALU or OBU based!", even though the track *is* OBU based.

This is because the check for the ISOM media subtype is erroneously skipped when dumping to stdout.

Fix it by doing the ISOM media subtype check earlier.


jeanlf:
thanks !

PR #2437, opened 2023-04-08
os_thread: Fix compiling with clang mingw

Biswa96:
This matches the return variable type with function prototype.
    Otherwise, clang in mingw shows the following error

    os_thread.c:225:10: error: incompatible pointer to integer
    conversion returning 'void *' from a function with result type
    'DWORD' (aka 'unsigned long') [-Wint-conversion]


aureliendavid:
all good thanks

PR #2438, opened 2023-04-08
support setting file extension for segments

kernrj:
This adds support for setting the file extension for init and regular segments.


rbouqueau:
Hi, I have the feeling this is the same as ```:segext=xxx``` and ```:initext=yyy``` e.g. ```gpac -i input -o output.mpg:segext=xxx:initext=yyy```. Do you confirm?

If so, what could we improve for people to find this more easily? What did you try prior to implementing it yourself?

kernrj:
We have a use case to set the file extension differently for audio and video CMAF files. I read through the wiki, looked at gpac -hh, and looked through the source and didn’t see an obvious way to do it. Setting segext and initext is for all segments, right? Is there a better way to set it per rendition?On Apr 8, 2023, at 12:51 PM, Romain Bouqueau ***@***.***> wrote:﻿
Hi, I have the feeling this is the same as :segext=xxx and :initext=yyy e.g. gpac -i input -o output.mpg:segext=xxx:initext=yyy. Do you confirm?
If so, what could we improve for people to find this more easily? What did you try prior to implementing it yourself?

rbouqueau:
It seems that even though the ```segext``` and ```initext``` properties could be inherited from the source, two factors prevent it:
1. The #Extension built-in property is 'D'/dropped if set on the source ([doc](https://github.com/gpac/gpac/wiki/filters_properties)). Anyway the dasher only considers this property, not the ```#MIMEType``` so it would fail with a custom file extension (as it would not recognize the type of the source).
2. I see this message ```Filter ffenc:segext option set but not used``` and ```segext``` / ```initext``` are not inherited from the source anyway. @jeanlf any idea why? Does it mean that the option is consumed and not forwarded anymore?

I see a clear use-case for this (CMAF with video/audio/sub extensions being resp. .cmfv/.cmfa/.cmfs). @kernrj Is your use-case different?

kernrj:
Exactly, we're using .cmfv and .cmfa for video and audio.

jeanlf:
looks good to me, could you just remove the man page from the PR ?

jeanlf:
thanks !

PR #2442, opened 2023-04-19
Fix infinite loop in configure with--disable-all

gorinje:


gorinje:
[fix-disable-all.patch](https://github.com/gpac/gpac/files/11277726/fix-disable-all.patch)

I've also attached some modification of the code to make --disable-all compiling again (it's difficult though to be sure that these are the the right corrections)

jeanlf:
thanks - we'll look at the disable flags checking in our CI

PR #2451, opened 2023-04-29
uncvg: fix bad pixel map value parsing

bradh:
As written, it uses `x` without defining it, and it drops the final column digit since `slice(begin,end)` does not include `end`.

I'm not sure I completely understand the expected format, but assuming it is something like:

```
./bin/gcc/gpac uncvg:vsize=1000x2000:bpm=r300,r350,c200,c240,c280,100x120,130x150:fps=0 -o ~/uncompressed_sbpm.heif
```

then this PR should fix the parsing.

(also, I see the sbpm box included even if it is the bpm is not specified in the filter. Is that intended?)


jeanlf:
Thanks !

PR #2458, opened 2023-05-08
dasher: put default precision to ms (period_start.den=1000 instead of 1)

rbouqueau:
... to avoid duration being truncated (and last segment missing)


rbouqueau:
Any news here?

jeanlf:
could you also do the same in context reloading: 
https://github.com/gpac/gpac/blob/master/src/filters/dasher.c#L5751

PR #2465, opened 2023-05-10
Fix HEVC SPS parsing

kernrj:
HEVC SPS parsing fails when sps_sub_layer_ordering_info _present_flag is set, and sps_max_sub_layers_minus1 is greater than 0. The result was three arrays weren’t being fully read.


jeanlf:
thanks!

PR #2499, opened 2023-06-24
Add nix flake to help builds with nix packages

RodolpheFouquet:
Hello,

This PR exists to add a nix flake in the repository to build GPAC (without the extra deps).

It integrates with https://nixos.org/ and allows you to build gpac (and get its dependencies) using ```ǹix build``` or just spawn an isolated development shell using ```nix develop```.

So far, only 64bit linux is supported, but mac and windows support could be enabled in the future. As well as 32bit linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,6 @@ gpac.pc
 tests/external_media/
 tests/results/
 tests/hash_refs/
+
+# nix output
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,25 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1687130670,
+        "narHash": "sha256-VKTdfsJe7sVTTqxTd3eRGPoUgEeJKD+kwS86B6TY874=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "c1bca7fe84c646cfd4ebf3482c0e6317a0b13f22",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "root": {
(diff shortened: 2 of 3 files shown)


aureliendavid:
Hi, 

That's very interesting. I don't know the nix project but I'm going to read more about it. 

I was in the process of setting up docker images to easily get a gpac build/runtime environment (see https://github.com/aureliendavid/gpac/blob/docker/build/docker/ubuntu.Dockerfile) 

But if I understand correctly this is a different approach from containerization, it uses environment variables to set up the dev environment (kinda like virtualenv+pip does for pyton). Which can be pretty nice, and maybe it can be useful in conjunction with docker. 

I'll explore a bit more but thanks for the PR!

aureliendavid:
Just tested it, it works pretty well. 

I think we can merge this as it can be useful to some. 

Just a question: is the lockfile necessary in the repo? 


Just to keep a trace of how to use it, here's what I did:

```shell
# install nix
sh <(curl -L https://nixos.org/nix/install) --no-daemon

# config
mkdir -p ~/.config/nix  ; echo "experimental-features = nix-command flakes" >> ~/.config/nix/nix.conf

# build (in gpac dir)
nix build

# use 
./result/bin/gpac

# enter dev env
nix develop
# do stuff

# exit dev env
exit
```

RodolpheFouquet:
Hi, sorry for the late reply.

Just a question: is the lockfile necessary in the repo?

=> I am relatively unsure, in theory, yes, as it should act like a .lock file with npm.

I would also like to find a way to create a version of the flake that builds the extra-deps to have a feature complete version of GPAC with nix flakes.

RodolpheFouquet:
By the way, instead of running `./result/bin/gpac` you can do

`nix shell --command gpac`
`nix shell --command MP4Box`

PR #2512, opened 2023-07-03
Add a NULL check of field.far_ptr in gf_dump_vrml_simple_field()

XidianGeneral:
In the commit c5249ee4b62d, `field.far_ptr` is checked before being used. However, the following code statement misses a NULL check of this variable: 
`list = * ((GF_ChildNodeItem **) field.far_ptr);` in `gf_dump_vrml_simple_field()`.
Thus, a NULL check should be added before this code statement.


jeanlf:
the access to field.far_ptr result to NULL list if field.far_ptr is NULL so `while`loop being skipped.

PR #2540, opened 2023-07-25
Correct check for 'modificationTime'

mjunix:


jeanlf:
thanks, I patched this as well as gf_isom_get_track_creation_time

PR #2555, opened 2023-08-14
Fix strnicmp lens on some fileimport comparisons

joedrago:
I was trying to use `colormx` overrides today using mp4box and it refused to be parsed on the commandline. I dug in and noticed that these string comparisons were checking the wrong lengths, so I figured I'd contribute the few I noticed to be wrong. There might be others hiding in there (this fixed my `colormx` usage, at least)!

While I'm at it, thanks for mp4box; it is great stuff.


rbouqueau:
Thank you Joe for the contribution and the nice words. All the best to you!

PR #2595, opened 2023-09-13
Fix nix flake for macOS

niklaskorz:
1. Generate outputs for all systems through flake-utils (x86_64 and aarch64 linux and darwin)
2. Add Carbon framework as build input on darwin (required by https://github.com/gpac/gpac/blob/d5dd620add58df9546d58211975943434364c060/applications/gpac/carbon_hook.c#L27)
3. Make some build inputs Linux specific: X11, mesa, alsa, jack2, pulseaudio
4. Move git to native build inputs (only relevant for cross compiling; native build inputs are specific to the host system, build inputs for the target system)
5. Only generate the static library `libgpac_static.a` if `STATIC_MODULES` is set. Nix prefers dynamic libraries (and actually passes a `--disable-static` to configure by default) and the static library adds unnecessary bloat. Furthermore, the presence of the static library seems to break the final binary (both gpac and MP4Box) in the Nix flake output for reasons I don't know yet (crashes immediately with `[1]    4966 killed     result/bin/gpac`), but with the static library disabled, everything is fine.
6. Remove the custom build phases as the default ones defined by Nix do the same (as long as `enableParallelBuilding = true` is defined).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,30 @@
 {
   "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1694529238,
+        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "ff7b65b44d01cf9ba6a71320833626af21126384",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1686960236,
-        "narHash": "sha256-AYCC9rXNLpUWzD9hm+askOfpliLEC9kwAo7ITJc4HIw=",
+        "lastModified": 1694422566,
+        "narHash": "sha256-lHJ+A9esOz9vln/3CJG23FV6Wd2OoOFbDeEs4cMGMqc=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "04af42f3b31dba0ef742d254456dc4c14eedac86",
+        "rev": "3a2786eea085f040a66ecde1bc3ddc7099f6dbeb",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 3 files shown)


niklaskorz:
🦗

RodolpheFouquet:
Hi I will have a look during the week end, I will tap into flake-utils and see if I can make a nix-darwin variant.

RodolpheFouquet:
First @niklaskorz, thank you very much for your contribution.

Hi, I will have a look later this week. I do not have access to the Mac right now.

> Only generate the static library libgpac_static.a if STATIC_MODULES is set. Nix prefers dynamic libraries (and actually passes a --disable-static to configure by default) and the static library adds unnecessary bloat. Furthermore, the presence of the static library seems to break the final binary (both gpac and MP4Box) in the Nix flake output for reasons I don't know yet (crashes immediately with [1]    4966 killed     result/bin/gpac), but with the static library disabled, everything is fine.

Agreed, the default should remain building both libraries as many internal GPAC script (and gpac test infrastructure) relies on this behaviour. I think DISABLE_STATIC shoud be the option, and be set by the flake to only.

We could still use a nix build argument to reenable static if needed.

niklaskorz:
Can you clarify what these variables in `configure` are for:

https://github.com/gpac/gpac/blob/666daf2931168e1f7752c7d2b915b56684213750/configure#L92C13-L94

If static builds should be enabled by default but with the option to disable them, shouldn't be one of these variables (or all of them?) be set to `yes` by default? I'm fine adding another generic `static = yes` variable (which would then be set to `no` by `--disable-static`) if none of the existing variables is meant to control static library compilation.

rbouqueau:
@niklaskorz I guess ```--static-build``` should be sufficient.

aureliendavid:
> Can you clarify what these variables in `configure` are for:
> 
> https://github.com/gpac/gpac/blob/666daf2931168e1f7752c7d2b915b56684213750/configure#L92-L94

these variables control the following switches described in the configure help: 

```
  --static-build           link statically against libgpac but still allow dependencies to shared libraries (enable --static-modules)
  --static-bin             enable static linking of MP4Box and gpac only (will enable --static-build), disable all libraries not linkable statically.
  --static-modules         embed modules in libgpac rather than using dynamic library modules
```

so they all control a different thing than whether or not the static libgpac is built at all, that's why they are all set to `no` by default. 

I guess adding a `--disable-static` to completely disable building `libgpac_static.a` could make sense. (but it would also be easier to let it build as is and just ignore/remove the lib you don't want a posteriori)

rbouqueau:
What's the status here?

RodolpheFouquet:
Hi, I am sorry for the delay, I have reviewed the PR and it works, however:

- I think --disable-static should be handle properly, a new config param should be introduced in our config.mak to handle it. I do not want to steal your work, so you can use the branch I have just created https://github.com/gpac/gpac/tree/feature/improve-nix-flake to modify your branch
- we should mark static builds as unsupported on nix with Darwin. Nix uses glibtool and not the Apple libtool and supporting it requires more work (we can open an issue for that).

I have tested the PR on NixOS unstable and on MacOS sonoma with Apple silicon.

aureliendavid:
I think we can close this for lack of progress/consensus about whether there's actually an issue here. 

Feel free to reopen this or a new PR/issue if there is still an interest in something to fix about nix flake. 

Thanks

PR #2644, opened 2023-10-15
Buildbot hls tsb live2vod

rbouqueau:


jeanlf:
reworked to avoid list copy, should now be ok for dash and hls

PR #2648, opened 2023-10-15
fix failure and warning on GPAC_FIXED_POINT

mhfan:
Some obvious mistakes.

### DIFF
--- a/src/utils/path2d.c
+++ b/src/utils/path2d.c
@@ -166,7 +166,6 @@ GF_Err gf_path_add_line_to_vec(GF_Path *gp, GF_Point2D *pt) {
 GF_EXPORT
 GF_Err gf_path_close(GF_Path *gp)
 {
-	Fixed diff;
 	GF_Point2D start, end;
 	if (!gp || !gp->n_contours) return GF_BAD_PARAM;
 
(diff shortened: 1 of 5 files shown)


src/utils/path2d.c
```diff
@@ -175,8 +175,8 @@ GF_Err gf_path_close(GF_Path *gp)
 	end = gp->points[gp->n_points-1];
 	end.x -= start.x;
 	end.y -= start.y;
-	diff = gf_mulfix(end.x, end.x) + gf_mulfix(end.y, end.y);
-	if (ABS(diff) > FIX_ONE/1000) {
+	//diff = gf_mulfix(end.x, end.x) + gf_mulfix(end.y, end.y);
+	if (FIX_ONE < ABS(end.x) || FIX_ONE < ABS(end.y)) { // ABS(diff) > FIX_ONE/1000
```
jeanlf, line 179: I think using FIX_ONE here is wrong, we could end up not closing a path while the old code was. Could you replace with FIX_ONE/100 ?

mhfan: done

jeanlf:
thanks!

PR #2653, opened 2023-10-19
fix integer overflow issue in isomedia/box_code_base.c:413

### LINKED ISSUE #2652: Integer overflow issue in isomedia/box_code_base.c:413
[Y] I looked for a similar issue and couldn't find any.
[Y] I tried with the latest version of GPAC. Installers available at http://gpac.io/downloads/gpac-nightly-builds/
[Y] I give enough information for contributors to reproduce my issue

Description
There is a integer overflow issue in isomedia/box_code_base.c:413

System info
Ubuntu 22.04.2 LTS
GPAC-2.2.1

Build command
./configure --enable-sanitizer && make

crash command
/usr/local/bin/MP4Box -saf -diod poc

poc_file:
[poc.zip](https://github.com/gpac/gpac/files/13040344/poc.zip)

Crash output:
isomedia/box_code_base.c:413:29: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself

xiaoxiaoafeifei:
fix #2652

### DIFF
--- a/include/gpac/internal/isomedia_dev.h
+++ b/include/gpac/internal/isomedia_dev.h
@@ -1189,7 +1189,7 @@ typedef struct
 	u32 r_currentEntryIndex;
 	u32 r_FirstSampleInEntry;
 
-	s32 max_cts_delta;
+	u64 max_cts_delta;
 	//u32 sample_num_max_cts_delta;
 } GF_CompositionOffsetBox;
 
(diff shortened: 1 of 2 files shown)


aureliendavid:
thanks for the report and the fix

I prefer to fix the specific problem where it occurred instead of touching something as generic as the ABS() macro (see fix in the original issue #2652) but if the problem appears in a few other locations I might use your suggestion as a more generic fix

PR #2709, opened 2023-12-29
typos, touching the API

rbouqueau:
I think it is ok though still not many people are using the API still. Better than leaving a typo in the long term imho.


jeanlf:
thanks !

PR #2719, opened 2024-01-09
Fixing some bugs in netcap filtering

touatily:
Fixing some bugs in netcap filtering:

- Accepting value `FF` when patching a byte in a packet
- Do not stop at the first netcap filtering rule


jeanlf:
thanks !

Issue #2717, opened 2024-01-08
Endless While Loop Due to Unupdated Variable in `else` Block


touatily:
https://github.com/gpac/gpac/blob/7aef8038c6bdd310e65000704e39afaa0e721048/src/media_tools/route_dmx.c#L1454

In this `else` block, the variable `payload`, which is involved in the `while` condition, is not updated. Consequently, we are stuck in an infinite while loop.

### Reproduction Scenario

1. The file used in this test:

https://github.com/gpac/gpac/assets/27826148/87c1287b-04a2-4bdb-8023-2256ff9d221d

2. Generating dash session:

```
gpac -i short.mp4  -o dash_session/live.mpd
```

3. Sending segments via ROUTE
```
gpac -i dash_session/live.mpd dashin:forward=mani -o route://234.1.1.1:1234/live.mpd -netcap=dst=route.gpc
```
4. Receiving the segments while patching the 13th byte of the first packet.

```
gpac -i route://234.1.1.1:1234/live.mpd -netcap=src=route.gpc,nrt,[s=1,o=12,v=0] inspect
```

### Result:

![image](https://github.com/gpac/gpac/assets/27826148/4f0c41e2-8428-486e-9666-4bf880349a0f)

aureliendavid:
We can add a `payload += i;` in the else to fix the immediate issue (infinite loop). 

But I think this whole section could use some more checks on string lengths and the like (what happens if a header doesn't end with a proper CRLF? or if the payload is truncated in the middle of a header? etc.) 

@touatily if you have time and want to do a PR for this fix + some string boundaries checks in the function that'd be nice :) (otherwise I'll have a look when I can)

PR #2723, opened 2024-01-10
Fixing issue 2717

touatily:
Fix issue https://github.com/gpac/gpac/issues/2717

- Verifying that payload is not erroneously treated as plaintext => print warning message
- Checking if end of payload is reached
- When header entity is unrecognized, avoid printing whole payload
- Print error message if end of payload is reached prematurely


aureliendavid:
(please wait before merging, a review is in progress in private)

aureliendavid:
thanks

PR #2724, opened 2024-01-10
Fixed ddts_box_read to position bitstream to end of ddts box, as expected by callers gf_isom_box_array_read and gf_isom_box_parse_ex.

Roy-Funderburk:
After calling ddts_box_read, gf_isom_box_array_read expects the bitstream to have been positioned to the end of the 'ddts' box as it will call gf_isom_box_parse_ex on any remaining box data.  My apologies for missing this in my prior submission.

### DIFF
--- a/src/isomedia/box_code_base.c
+++ b/src/isomedia/box_code_base.c
@@ -12450,6 +12450,7 @@ void ddts_box_del(GF_Box *s)
 GF_Err ddts_box_read(GF_Box *s, GF_BitStream *bs)
 {
 	GF_DTSSpecificBox *ptr = (GF_DTSSpecificBox *)s;
+	const u64 ddtsBoxSize = 20;
 
 	ptr->cfg.SamplingFrequency = gf_bs_read_u32(bs);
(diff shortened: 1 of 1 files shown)


src/isomedia/box_code_base.c
```diff
@@ -12450,6 +12450,7 @@ void ddts_box_del(GF_Box *s)
 GF_Err ddts_box_read(GF_Box *s, GF_BitStream *bs)
 {
 	GF_DTSSpecificBox *ptr = (GF_DTSSpecificBox *)s;
```
jeanlf, line 12452: I forgot that in the prev review: box parsing code should use the macro ISOM_DECREASE_SIZE when reading, so your code should just be (preferably before first gf_bs_read):
ISOM_DECREASE_SIZE(ptr, 20)

you then do not need to check if there are remaining bytes, this is done by the caller

Roy-Funderburk: I uploaded a new commit with the suggested changes, and that looks a lot nicer.  Thanks!

jeanlf:
thanks - I'll update the udts_box_read for alloc safety

PR #2728, opened 2024-01-14
Fixed track duration calculation in -edits

saindriches:
When `movie_ts` and `media_ts` are not consistent, the current code may lead to issues, potentially resulting in a shorter `edur` than expected.

### Reproduction
Generate an MP4 file with a duration of 10 seconds, where both the video timescale and framerate are set to 50.
```bash
ffmpeg -lavfi avsynctest=framerate=50 -video_track_timescale 50 -c:v libx264 -bf 0 -t 10 test.mp4
```
Use edit lists to skip the first 5 seconds.
```bash
MP4Box -edits 1=re0-5,221524/44100 -edits 2=re0-5,5 test.mp4
```
The MP4 file should have a timescale of 1000, so the track duration in the edit lists should be 5000. For audio, it's okay because the audio timescale is larger (44100). However, for video, the duration is just 500 in video timescale, then the video will freeze near 0.5s.

On the other hand, according to my understanding of the original code, if there is a need to clamp `edur`, it should be clamped to the actual available duration based on `media_time`.

### DIFF
--- a/applications/mp4box/fileimport.c
+++ b/applications/mp4box/fileimport.c
@@ -589,7 +589,7 @@ GF_Err apply_edits(GF_ISOFile *dest, u32 track, char *edits)
 			if (e) goto error;
 		}
 		else if (edits[0]=='e') {
-			u64 movie_t, media_t, edur;
+			u64 movie_t, media_t, edur, edur_max;
 			u32 rate;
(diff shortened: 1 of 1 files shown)


jeanlf:
thanks for the report, now fixed on master , including adjusting media dur with media start time

PR #2738, opened 2024-01-23
net: fix multicast IPV6 hops default value

AlaricSenat:
Forcing the HOPS to zero when not specified is quite confusing to the user. We let the IPV4 implementation to default when TTL isn't specified, let's make IPV6 have the same behavior.


jeanlf:
Thanks !

PR #2739, opened 2024-01-25
Fix discarding manifests

touatily:
- Corrected some typos.
- Fix: to prevent discarding future versions of the manifest if the first one is not correctly received.


jeanlf:
thanks

PR #2740, opened 2024-01-26
Improved console experience in gpac wasm demo page

DenizUgur:
This PR imrproves the console experience by these changes:
- Taller console
- ANSI color codes shown properly
- Some useful shortcuts
    - `Ctrl+C`: Clear current command
    - `Ctrl+L`: Clear console
    - `Ctrl+A`: Move cursor to beginning of command
    - `Ctrl+E`: Move cursor to end of command

PR #2745, opened 2024-01-29
WASM demo page bug fixes

DenizUgur:
This is related to #2740. I've realized there were some problems with it. They are fixed by preventing default shortcuts on document level.


jeanlf:
thanks

PR #2748, opened 2024-02-01
Buildbot scte35 id3 markers

rbouqueau:
- M2TS demux:
  - [DONE] Handle metadata in PES and Sections as it was done for TEMI in the AdaptationField.
  - [DONE] ID3/meta handling shall be more generic for proper inspection.
  - [DONE] Convey as a property.
  - [TODO] Make sure there is no duplication between the PID (when section or PES) and properties. It may be problematic when remuxing.
  - [TODO] Make configurable which TS metadata track goes into which CMAF track, for instance, based on the TS PID.
- Inspect:
  - [DONE] generic dump of all SCTE-35 commands
  - [DONE] detail for splice_time/time_signal and splice_insert (which are the only two commands in my samples).
  - [TODO] test dump from timed metadata track: https://dashif-documents.azurewebsites.net/Ingest/master/DASH-IF-Ingest.html#interface-1-timed-metadata. SAMPLE NEEDED. 
- M2TS mux:
  - [TODO] Re-add SCTE PID from property.
- Dasher/isom_mux:
  - [TODO] SCTE-35: add ```<InbandEventStream schemeIdUri="urn:scte:scte35:2013:bin"></InbandEventStream>```.
  - [TODO] the current implementation starts a fragment when an 'emsg' box is inserted because it shall be right before the 'moof' box. Add option to wait for the next fragment.
  -  [TODO] several messages in the same 'emsg' box (i.e. same message_data[] array) shall be separated with '0x0A'.
  - [TODO] when the duration is unknown the event_duration should be set to 0XFFFFFFFF.
  - [TODO] investigate brands and URNs.
  - [TODO] Add HLS ```#EXT-X-CUE-IN``` and ```#EXT-X-CUE-OUT```.
- Metadata injection new module (likely jsf):
  - [TODO] Insert timecodes as a property. Then reinsert the AV1/metadata in ```rfobu``` or the AVC/HEVC SEI in ```rfnalu```. Timecode in HLS to be inserted as ```#EXT-X-PROGRAM-DATE-TIME```.
  - [TODO] Insert SCTE-35.
  - [TODO] Insert KLVs (which translates into emsg boxes). Specify each track


src/filters/inspect.c
```diff
@@ -2157,6 +2157,169 @@ static void dump_temi_time(GF_InspectCtx *ctx, PidCtx *pctx, FILE *dump, const c
...
+	} else {
+		inspect_printf(dump, " SCTE35");
+	}
+
+	if (!pctx->bs)
+		pctx->bs = gf_bs_new(att->value.data.ptr, att->value.data.size, GF_BITSTREAM_READ);
+	else
+		gf_bs_reassign_buffer(pctx->bs, att->value.data.ptr, att->value.data.size);
+
+	uint8_t table_id = gf_bs_read_u8(pctx->bs);
```
aureliendavid, line 2184: `uint8_t` currently doesn't build on windows

maybe we should add an `#include <stdint.h>` in the windows part of `setup.h`?

or maybe we should just use the `u8` type instead?

rbouqueau: I will use u8, thanks! FYI this is just a draft PR ATM. I'm not even sure we will not cherry-pick the commits on here.

src/filters/dmx_m2ts.c
```diff
@@ -1106,12 +1168,86 @@ static void m2tsdmx_on_event(GF_M2TS_Demuxer *ts, u32 evt_type, void *param)
...
+			if (p->value.uint != GF_STREAM_AUDIO)
+				continue;
+
+			GF_SAFEALLOC(t, GF_M2TS_Prop);
+			if (!t) break;
+			t->type = M2TS_ID3;
+			bs = gf_bs_new(NULL, 0, GF_BITSTREAM_WRITE);
+			gf_bs_write_u32(bs, 90000);                     // timescale
+			gf_bs_write_u64(bs, pck->PTS);                  // pts
+			gf_bs_write_data(bs, pck->data, pck->data_len); // data
```
jadarve, line 1197: Does it make sense to write `pck->data_len` before `pck->data` so that other filters that parse this data know how many bytes to read? For instance [here](https://github.com/gpac/gpac/pull/2748/files#diff-e0502e3bb4133c2be5d6f040521b18af03378ede1e3dfb4c7c71910ef2a7136cR6349)

rbouqueau: We know the size of the property. That being said what would happen if several properties are pushed simultaneously? The spec says that "several messages in the same 'emsg' box (i.e. same message_data[] array) shall be separated with '0x0A'." Is it a realistic use-case?

PR #2752, opened 2024-02-08
Fix: Do Not Skip Overlapping LCT Fragments

touatily:
When receiving LCT fragments, ensure that packets are not skipped if they overlap with received packets. 
This fix addresses the issue where skipping overlapped packets led to potential data loss or processing errors.

### DIFF
--- a/src/media_tools/route_dmx.c
+++ b/src/media_tools/route_dmx.c
@@ -23,6 +23,7 @@
  *
  */
 
+#include <math.h>
 #include <gpac/route.h>
 
 #if !defined(GPAC_DISABLE_ROUTE)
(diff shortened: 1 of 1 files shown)


src/media_tools/route_dmx.c
```diff
@@ -23,6 +23,7 @@
  *
  */
 
+#include <math.h>
```
rbouqueau, line 26: You could remove this by using ```MAX``` instead of ```fmax```.

rbouqueau:
I made 2 minor comments. Otherwise LGTM. Thanks!

rbouqueau:
Thanks for the commits, they address my comments.

PR #2765, opened 2024-02-27
Enhancements for Manifest Parsing and Validation

touatily:
-  Replace sscanf with strtol to ensure TOI field is parsed as an integer.
-  Replace sscanf with strtol to ensure TSI field is parsed as an integer.
-  Validate that the dport field is an integer within the range [0, 65535].
-  Verify if session channels are defined in the manifest.


aureliendavid:
seems all good to me, thanks for the contrib!

PR #2779, opened 2024-03-12
Fix bugs found by fuzzer

touatily:
- Fixing runtime error: 
> negation of -2147483648 cannot be represented in type 'int'

- Check if `filter` field is not null also to prevent segmentation fault
- In Route dmx, update `alloc_size` of an object if necessary when picking it from reservoir
- Cast to `u64` to avoid overflow in comparisons
- Purge segment if there are any fragments that extend beyond the total length
- Display warning message if received TOL is null


aureliendavid:
looks good, thanks!

PR #2780, opened 2024-03-12
use plain memcpy instead of zmemcpy macro

rathann:
This fixes build with zlib-ng, which doesn't define the macro.

Regardless, this macro was defined to memcpy in most cases anyway: https://github.com/madler/zlib/blob/master/zutil.h#L209


jeanlf:
thanks !

PR #2788, opened 2024-03-22
Error handling fixes

touatily:
Fix some autofuzz bugs:

- `nJauzoDhd5DctA`: skip packet when `start_offset + size > GF_UINT_MAX`  (Not supported)
- `yXRaBOYUxQRZFA`: null pointer passed as  second argument to `sscanf`


aureliendavid:
thanks!

PR #2792, opened 2024-04-02
Fix infinite loop when segment_template is absent in MPD

touatily:
This pull request addresses the issue of an infinite loop encountered when executing the specified command line.

### Reproduction Steps:

- Download the capture file [route_nozip.zip](https://github.com/gpac/gpac/files/14836928/route_nozip.zip).
- Execute the following command line:
```
gpac -i route://234.1.1.1:1234/live.mpd -netcap=src=route_nozip.gpc,nrt,[s=1,o=847,v=40] inspect -logs=route@info
```

### Expected Behavior:

The command should execute without encountering an infinite loop.

**_Note_:** I am not sure if this is the correct approach to fix the issue, which is why I created this as a draft PR.


jeanlf (approved):

jeanlf:
thanks!

PR #2803, opened 2024-04-14
fix time source for wasm

### LINKED ISSUE #2775: WebAssembly: Add threading for download manager
For some reason, download manager is not working properly in a worker. Considering both the main thread and worker thread are running on their own thread, I can't find the reason why download manager experiences degraded performance while using gpac in a worker.

What happens is that when using worker and a remote source, gpac takes almost 60-70 times more to load the source. It seems that the reason lies with scheduling `fetch` promises while the same thread runs gpac. I've tried a very rough way to run `fetch` in seperate threads but I think I'm still blocked by scheduling. The correct way to achieve this seems to be through native (pthread) threading for download manager. Just how `-dm-threads` option supposed to work.

I would be happy to open a PR to bring `-dm-threads` to WebAssembly version but I need some guidance on what is needed and what might be a blocking issue. @jeanlf any idea on how to proceed with this one?

**My attempt at multi-threaded fetch:**
```js
let workerScript = `self.onmessage = async function(event)
{
	let {url, options} = event.data;
	let res = await fetch(url, options);
	if (res.ok)
	{
		self.postMessage({
			ok: res.ok,
			url: res.url,
			status: res.status,
			headers: JSON.stringify([...res.headers]),
			redirected: res.redirected,
			body: res.body
		}, [res.body])
	} else self.postMessage({ok: res.ok, status: res.status})
}`;
let workerBlob = new Blob([workerScript], { type: 'application/javascript' });
fetcher._worker = new Worker(URL.createObjectURL(workerBlob));

fetcher._worker.onmessage = (event) =>
{
	if (event.data.ok)
	{
		fetcher._state = 1;
		fetcher._bytes = 0;
		fetcher._reader = event.data.body.getReader();

		let final_url = null;
		if (event.data.redirected) final_url = event.data.url;
		libgpac._fetcher_set_reply(fetcher.sess, event.data.status, final_url);

		let headers = JSON.parse(event.data.headers);
		headers.forEach(([value, key]) => {
			libgpac._fetcher_set_header(fetcher.sess, key, value);
		});
		libgpac._fetcher_set_header(fetcher.sess, 0, 0);
	}
	else
	{
		libgpac._fetcher_set_reply(fetcher.sess, event.data.status, null);
		do_log_err('fetcher for ' + url + ' failed ' + event.data.status);
		fetcher._state = 3;
	}
};

fetcher._state = 0;
fetcher._worker.postMessage({url, options}, [...(options.body ? [options.body] : [])]);
return ret;
```

**Was placed here:**
https://github.com/gpac/gpac/blob/96b474ec59b3490eed7a874b14f78ad4b3a9d2bd/src/utils/downloader.c#L7702

DenizUgur:
In wasm build, usage of `gettimeofday()` is painfully slow when gpac runs on a worker. Emscripten offers a [substitute](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_get_now) for it. I had to make this change because when I profiled my application, almost all of the time wasm made a call to javascript to get the time and it stalled the entie stack because of it.

I haven't fully tested it but as emscripten documentation pointed out, this might use `performance.now()`. Which means between different workers (threads) this value will have a different origin ([mdn reference](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#performance.now_vs._date.now)). Depending on how these clocks are used internally the implementation might need to change. For reference, here is how this function is defined in emscripten:

https://github.com/emscripten-core/emscripten/blob/main/system/lib/standalone/standalone.c#L179-L185

fixes #2775
related https://github.com/urho3d/Urho3D/issues/916

### DIFF
--- a/src/utils/os_divers.c
+++ b/src/utils/os_divers.c
@@ -133,7 +133,7 @@ u32 gf_gpac_abi_micro()
 }
 
 
-#ifndef WIN32
+#if !defined(WIN32) && !defined(GPAC_CONFIG_EMSCRIPTEN)
 
 GF_EXPORT
 u32 gf_sys_clock()
@@ -149,6 +149,23 @@ u64 gf_sys_clock_high_res()
 	struct timeval now;
 	gettimeofday(&now, NULL);
 	return (now.tv_sec)*1000000 + (now.tv_usec) - sys_start_time_hr;
+}
+
+#endif
+
+#ifdef GPAC_CONFIG_EMSCRIPTEN
+#include <emscripten/emscripten.h>
+
+GF_EXPORT
+u32 gf_sys_clock()
+{
+	return (u32)(emscripten_get_now() - sys_start_time);
+}
+
+GF_EXPORT
+u64 gf_sys_clock_high_res()
+{
+	return (long long)(emscripten_get_now() * 1000.0) - sys_start_time_hr;
 }
 
 #endif


jeanlf:
According to the [emscripten worker doc](https://emscripten.org/docs/api_reference/wasm_workers.html#pthreads-synchronize-wallclock-times), using emscripten_get_now() should be time-aligned in pthreads, so I think we can merge. Thanks !

PR #2822, opened 2024-05-01
configure: Fix finding xvid by including stddef.h for definition of NULL

ryandesign:
The configure test for xvid was failing to find it due to

```
*** CC/CXX Test Failed (res 1 args -L/opt/local/lib -Wl,-headerpad_max_install_names -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -arch x86_64 -L/opt/local/lib -lxvidcore -lpthread) : 

/opt/local/var/macports/build/_Volumes_Shared_macports-ports_multimedia_gpac/gpac/work/.tmp/gpac-conf-30001-60580-19987.c:2:70: error: use of undeclared identifier 'NULL'
int main( void ) { void *codec; xvid_decore(codec, XVID_DEC_DESTROY, NULL, NULL); return 0; }
                                                                     ^
/opt/local/var/macports/build/_Volumes_Shared_macports-ports_multimedia_gpac/gpac/work/.tmp/gpac-conf-30001-60580-19987.c:2:76: error: use of undeclared identifier 'NULL'
int main( void ) { void *codec; xvid_decore(codec, XVID_DEC_DESTROY, NULL, NULL); return 0; }
                                                                           ^
2 errors generated.

Source was: 
#include <xvid.h>
int main( void ) { void *codec; xvid_decore(codec, XVID_DEC_DESTROY, NULL, NULL); return 0; }
```

Fix by including <stddef.h> anytime `NULL` is referenced.

### DIFF
--- a/configure
+++ b/configure
@@ -1591,6 +1591,7 @@ config_package ssl "openssl" "" "$ssl_lflags" "" '#include <openssl/ssl.h>
 #include <openssl/x509.h>
 #include <openssl/err.h>
 #include <openssl/rand.h>
+#include <stddef.h>
 int main( void ) {  SSL_CTX_set_options(NULL, SSL_OP_ALL); return 0; }'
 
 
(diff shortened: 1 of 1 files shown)


rbouqueau:
Thank you! LGTM, merging.

PR #2836, opened 2024-05-12
quickjs-libc: fix incompatible pointer error with gcc 14

1480c1:
gcc 14 promoted -Wincompatible-pointer-types to an error and now shows

```c
src/quickjs/quickjs-libc.c:3376:49: error: passing argument 2 of 'GetExitCodeProcess' from incompatible pointer type [-Wincompatible-pointer-types]
 3376 |                 GetExitCodeProcess(pi.hProcess, &ret);
      |                                                 ^~~~
      |                                                 |
      |                                                 int *
```

### DIFF
--- a/src/quickjs/quickjs-libc.c
+++ b/src/quickjs/quickjs-libc.c
@@ -3291,7 +3291,7 @@ static JSValue js_os_exec(JSContext *ctx, JSValueConst this_val,
 	char *exec_argv = NULL, *file = NULL, *cwd = NULL;
 	char *envp = NULL;
 	uint32_t exec_argc, i;
-	int ret;
+	DWORD ret;
 	BOOL block_flag = TRUE, use_path = TRUE;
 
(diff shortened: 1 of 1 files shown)


rbouqueau:
Thank you. Actually the modified code was copied from quickjs so it would seem logical to also modify the upstream if not done already.

I checked and it seems quickjs is available and actively maintained on a git repository at https://github.com/bellard/quickjs/. Maybe GPAC could change the way it integrates quickjs?

1480c1:
I believe that the changes here do not apply to upstream as they do not use `GetExitCodeProcess()`, and thus they don't encounter the same issue. Rather, it seems like these changes are due to the mentioned `- added MSVC support for most 'os' functions (most importantly exec, waitpid and kill)` feature specific to this version of quickjs

jeanlf:
Thanks, merging - we'll update QuickJS to latest version soon

PR #2853, opened 2024-05-30
Dockerfile for wasm build

DenizUgur:
I've also removed the `build/` entry in `.gitignore` as I think that might have been a mistake.

The relevant wiki changes are here: gpac/wiki#6

- [x] Copy the gpac repository locally
- [ ] Fix Node path
- [x] Evaluate if autoconf is required again
- [x] Copy `gpac.html` without symlink


rbouqueau:
When building the container it seems that the node version is forced to 18.20.3 instead of 16.20.3 in the Dockerfile:

```
Resolving SDK version '3.1.56' to 'sdk-releases-9d106be887796484c4aaffc9dc45f48a8810f336-64bit'
Setting the following tools as active:
   node-18.20.3-64bit
```

It almost works but gpac.html location seems broken:
```
rbouqueau@rose:~/works/gpac/gpac$ ls -l /tmp/tmp/wasm
total 37948
lrwxrwxrwx 1 rbouqueau rbouqueau       39 May 31 06:04 gpac.html -> /gpac_public/share/emscripten/gpac.html
```

If I fix that manually it runs :)

DenizUgur:
Do we need it to be Node 16?

Using the commands mentioned in the wiki, I was able to run it without any changes. Have you followed that? Or what exactly did you had to change?

rbouqueau:
> Do we need it to be Node 16?

I don't think so. This seems inconsistent so I reported it here.

> Using the commands mentioned in the wiki, I was able to run it without any changes. Have you followed that? Or what exactly did you had to change?

No change and it worked for me!

DenizUgur:
> This seems inconsistent so I reported it here.

Ah I see what you mean. Weird that it still works. Maybe it doesn't even use that env variable.

rbouqueau:
@DenizUgur Are you good on this?

Are other people ok to merge?

nlsdvl:
I tested it on my hand. The EMSDK version used here is the same as used by the buildbot and compatible with the gpac.html page. 

Using a more recent EMSDK will break the current gpac.html and require updating it.

aureliendavid:
Hi, 

Thanks for the PR. 

One hesitation I had when writing the dockerfile for ubuntu was how to get the source into the image. There are two options: 

1. via git clone
2. via COPY

Both have their own pluses and minuses. 

Getting the source via git clone allows to use the dockerfile on its own, outside of the a working copy of the repo. 

However the drawback is that with this method you can't build dev changes directly, and you can't build branches and PRs (since you're always cloning the gpac master branch head). 

For the ubuntu dockerfile I wanted to have a github action that builds master and upload it to dockerhub. I also wanted to build branches that are pushed directly to the repo. But I didn't want to build all PRs automatically because it's a very easy thing to abuse.  (see https://github.com/gpac/gpac/blob/master/.github/workflows/ga.yml)

I also wanted the ability for users to dev on a local repo and build with docker. 

For these reasons I opted for solution 2. 

I wonder if we should do the same for wasm or if the git clone solution has a better justification here?

<hr >

Other small remarks: 

- why do we need to build the latest version of autoconf? is the debian one too old?
- from the wiki when we do `docker cp gpac-wasm:/gpac_public/bin/gcc/. /path/to/destination` there's an issue with `gpac.html` which is a symlink to `/gpac_public/share/emscripten/gpac.html` and docker cp doesn't have a good way to follow symlinks ; so in order for this to work well we could copy the actual file in the image after the build


<hr >

here are my proposed changes:

```diff
diff --git a/build/docker/wasm.Dockerfile b/build/docker/wasm.Dockerfile
index 98f6ce27f..adb7d878d 100644
--- a/build/docker/wasm.Dockerfile
+++ b/build/docker/wasm.Dockerfile
@@ -47,7 +47,7 @@ RUN ./wasm_extra_libs.sh --enable-threading
 FROM base as gpac
 
 # Checkout GPAC source code
-RUN git clone --depth 1 https://github.com/gpac/gpac.git gpac_public
+COPY . gpac_public
 WORKDIR /gpac_public
 
 # Copy GPAC WASM dependencies
@@ -55,8 +55,9 @@ COPY --from=deps /deps_wasm/wasm_thread /deps_wasm/wasm_thread
 ENV PKG_CONFIG_PATH=/deps_wasm/wasm_thread/lib/pkgconfig
 
 # Configure GPAC
-RUN ./configure --emscripten --extra-cflags="-Wno-pointer-sign -Wno-implicit-const-int-float-conversion"
+RUN make distclean ; ./configure --emscripten --extra-cflags="-Wno-pointer-sign -Wno-implicit-const-int-float-conversion"
 
 # Build GPAC
 RUN sed -i "s/-sMODULARIZE=1/-O2 -sMODULARIZE=1/g" applications/gpac/Makefile
 RUN make -j$(nproc)
+RUN cp -a --remove-destination /gpac_public/share/emscripten/gpac.html  /gpac_public/bin/gcc/
```

aureliendavid:
I'm also having some issues with the produced files. 

Here's how I tested: 

```bash
docker build -t gpac-wasm -f build/docker/wasm.Dockerfile .  # with or without my changes

docker create --name gpac-wasm gpac-wasm
docker cp gpac-wasm:/gpac_public/bin/gcc/. twasm
docker rm gpac-wasm

gpac -p=0 -logs=http@debug   httpout:port=8080:rdirs=./twasm:cors=on:cert=cert.pem:pkey=key.pem
```

it does run but among several issues the gpac commands in the browser never return. 

If I run `gpac -version` in the wasm command line it displays the result but it doesn't return and always displays "Running gpac gpac -version" in red above the console. 

If I take the files produced by the buildbot and serve them in the same way and run the same test it works properly. 

Another test that fail with the docker-produced files (and not with the buildbot files) is: 

```
$ gpac -i https://localhost:8080/test.mp4 vout

[Downloader] Failed to initialize fetch
Filter httpin failed to setup: Internal Service Error
Failed to load filter for "https://172.20.155.53:8080/test.mp4": Internal Service Error
```

(with test.mp4 being a sample video I copied alongside the wasm files) 


As an indication the file sizes seem pretty different: 

```bash
# docker files

$ ll twasm/
total 38M
-rw-r--r-- 1 adavid adavid  96K Jun  5 14:49 gpac.html
-rw-r--r-- 1 adavid adavid 697K Jun  5 17:30 gpac.js
-rwxr-xr-x 1 adavid adavid  26M Jun  5 17:30 gpac.wasm
-rw-r--r-- 1 adavid adavid 2.8K Jun  5 17:30 gpac.worker.js
-rw-r--r-- 1 adavid adavid  11M Jun  5 17:29 libgpac_static.a
drwxr-xr-x 2 adavid adavid 4.0K Jun  5 17:28 temp
-rw-r--r-- 1 adavid adavid  66K Jun  5 16:26 test.mp4


# buildbot files

$ ll th/
total 45M
-rw-r--r-- 1 adavid adavid  96K May 29 17:13 gpac.html
-rw-r--r-- 1 adavid adavid 738K Jun  5 16:01 gpac.js
-rw-r--r-- 1 adavid adavid  34M Jun  5 16:01 gpac.wasm
-rw-r--r-- 1 adavid adavid 7.4K Jun  5 16:01 gpac.worker.js
-rw-r--r-- 1 adavid adavid  11M Jun  5 16:00 libgpac_static.a
drwxr-xr-x 2 adavid adavid 4.0K Feb 22  2023 temp
-rw-r--r-- 1 adavid adavid  66K Jun  5 17:43 test.mp4
```

I haven't tested with apache hosting. 

Can you reproduce?

DenizUgur:
Hi @aureliendavid,

Thank you for all the feedback.

> I wonder if we should do the same for wasm or if the git clone solution has a better justification here?

Originally, I used this within anohter project and I wanted the docker file to be isolated. However, inside this repository probably it should use `COPY` instead.

> why do we need to build the latest version of autoconf? is the debian one too old?

I believe so. IIRC without it some of the dependencies (deps_wasm) didn't build

> docker cp doesn't have a good way to follow symlinks

It worked for me but you are right. No need to take chanes here. I'll fix that.

I'll also look at your last comment right now.

DenizUgur:
> Can you reproduce?

All of the things you mentioned is caused by js glue code not being able to access `libgpac` reference.

The only way to fix it without breaking buildbot is by making a change in the `gpac.html`

```diff
<script>
  window.addEventListener('load', function() {
+   Object.assign(libgpac, LIBGPAC);
    libgpac(LIBGPAC);
  });
</script>
```

We could probably just assign the necessary properties but why make it more complicated than it is?

DenizUgur:
Actually, this doesn't seem to be enough. Many functions still has a reference to `libgpac`. They can be fixed by removing the reference but that would be a breaking change for the buildbot.

https://github.com/gpac/gpac/blob/c3419d92a35c0a98f41c251d83b885c1999fc311/src/utils/downloader.c#L7390-L7391

Our only option seems to be downgrading emscripten version to buildbot's if we can't/wont update buildbot.

aureliendavid:
> Our only option seems to be downgrading emscripten version to buildbot's if we can't/wont update buildbot.

I think we should do that anyway. The docker should probably reflect the "official" build of the buildbot (and the version we're developing with) and when we're ready we will upgrade both, no?



> I believe so. IIRC without it some of the dependencies (deps_wasm) didn't build

It seems to work without it for me. 

This brings another remark. You're using `debian:bullseye-slim` as a base image. This seem pretty old (bullseye is oldstable, and debian is pretty slow to upgrade stuff). 

For debian I think using the channels (stable, testing, etc.) might be better than specific version-codenames. 

`debian:testing-slim` for example would be pretty up to date (or `stable-slim` if we want to be conservative) 

we could even use `ubuntu:latest` to be closer to the buildbot build (and to the linux docker build). It looks like the ubuntu base image is not bigger than the debian -slim images (but I don't really have a preference between the two).

DenizUgur:
What is the buildbot Emscripten version? I couldn't figure it out from the logs.

I can remove autoconf stuff and change the base image to Ubuntu. I don't have a preference on either one of them.

aureliendavid:
If i'm not mistaken it is 3.1.31

```
$ ./emsdk list

The *recommended* precompiled SDK download is 3.1.31 (1eec24930cb2f56f6d9cd10ffcb031e27ea4157a).

To install/activate it use:
         latest

This is equivalent to installing/activating:
         3.1.31             INSTALLED
```

aureliendavid:
sorry it's actually 3.1.32 with node-14.18.2-64bit

DenizUgur:
Okay 3.1.31 is working. I'll try with 3.1.32 and push the changes.

DenizUgur:
We could merge this if this is satisfactory for you @aureliendavid

aureliendavid:
I'll retest tomorrow and merge if it's all good

once it's merged i'll also add the build to the github actions and upload to dockerhub 

thanks for the contribs!

aureliendavid:
Hello again, 

I still had issues with the new dockerfile but I finally understood why. 

The build produced a 10MB gpac.wasm (which is way too small) and wasn't able to open a simple mp4 file. 

The reason is we don't build ffmpeg. 

There was an issue with the dependency script that didn't report errors properly so the ffmpeg build error didn't cause a docker build fail like it should. I fixed the script in https://github.com/gpac/deps_wasm/commit/301480c4eb541dda2c8386d52b70b38b01cdd58f 

Now the reason for the ffmpeg build fail is because it didn't find libx265, even though x265 built properly. 

The reason ffmpeg didn't find libx265 is because the x265 build didn't produce .pc files for pkg-config. 

The reason x265 didn't produce these files is because the cmake instruction to do so is inside a "if (X265_LATEST_TAG)" condition

This variable is set by looking at the git tags of the cloned x265 repo. 

But since we're using a shallow clone here we don't have git tags so we don't have the variable so we don't have the pkg-config file so we don't have ffmpeg so gpac.wasm is bugged. 

Isn't this fun? (no)

Anyway the easiest solution is to remove the "--shallow-submodules" from the command line and it all seems much better. 

<hr>

```diff
diff --git a/build/docker/wasm.Dockerfile b/build/docker/wasm.Dockerfile
index b70d8b7ca..78e3669b2 100644
--- a/build/docker/wasm.Dockerfile
+++ b/build/docker/wasm.Dockerfile
@@ -30,7 +30,7 @@ ENV EMSDK_NODE="/emsdk/node/18.20.3_64bit/bin/node"
 FROM base as deps
 
 # Install GPAC WASM dependencies
-RUN git clone --recurse-submodules --shallow-submodules --depth 1 https://github.com/gpac/deps_wasm
+RUN git clone --recurse-submodules --depth 1 https://github.com/gpac/deps_wasm
 
 # Build GPAC WASM dependencies
 WORKDIR /deps_wasm
```

DenizUgur:
Wow this is interesting. I haven't encountered this probably because x265 doesn't even build in newer versions of Emscripten.

DenizUgur:
Confirmed, it's transcoding as expected. Thank you for the investigation.

aureliendavid:
seems ok now

will update when it's integrated with github actions and dockerhub

DenizUgur:
I have a existing GitHub workflow that checks the refs and invalidates the docker workflow cache. It's working pretty good. I can open a PR for you to look at

DenizUgur:
I agree, it would be better if `(ubuntu|wasm)_deps` uploaded their own image as well. Caching the docker build would only help for emscripten layer after that.

```yaml
- name: Setup Docker
  uses: docker/setup-buildx-action@v3

- name: Cache Docker layers
  uses: actions/cache@v4
  with:
    path: /tmp/.buildx-cache
    key: ${{ runner.os }}-buildx-${{ github.sha }}
    restore-keys: ${{ runner.os }}-buildx-

- name: Build the gpac WebAssembly target
  uses: docker/build-push-action@v5
  with:
    tags: gpac-wasm
    cache-from: type=local,src=/tmp/.buildx-cache
    cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max

- name: Move the cache to the correct location
  run: |
    rm -rf /tmp/.buildx-cache
    mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```

I'd be happy to help on that matter by the way. Each `*deps` repository would upload an image and trigger a build in this repository when that happens.

PR #2854, opened 2024-05-31
Fix `pthread_mutex_destroy failed` issue during repair process

touatily:
This PR addresses the error message `pthread_mutex_destroy failed with error code 16` during the repair process.

**Changes:**

- In the function `routedmx_check_blob_range`, unlock mutex `blob->mx` before returning.


jeanlf:
thanks!

PR #2857, opened 2024-06-03
Fix parents of moov,cmov,cmvd in box_funcs

y-guyon:
See context at
https://github.com/AOMediaCodec/libavif/pull/2182


y-guyon:
This requires many further changes to compile.  
Abandoning, even though displaying `<MovieBox Container="file cmvd">` is misleading for an AVIF file.

y-guyon:
Adapting `gf_isom_box_dump_start_ex()` to print the actual box parent instead of the possible containers would be another way to go:

https://github.com/gpac/gpac/blob/b779331e5b598f7c09f1f2d4943c63e695a7250a/src/isomedia/box_funcs.c#L2283

but `gf_isom_box_dump_start()` does not provide this context and is called too many times.

jeanlf:
I patched the container code for cmov/cmvd which were indeed wrong. Note that `MP4Box -diso` will not show cmof/cmvd as these are lost when we load the file; they can only be seen when activating logs for `container@debug`.

PR #2860, opened 2024-06-06
fix: runtime error member access within null pointer

touatily:
Fix runtime error: member access within null pointer
Add check on `movie->moov` variable

autofuzz bug DmhCZI7TfPXPxQ


aureliendavid:
thanks!

PR #2864, opened 2024-06-07
Support for the Multiview extension of H.265/HEVC

matteovn:
- A fix for the parsing of the SPS is provided. This is because for NALUs associated with layers whose identifier is greater than 0 the spec requires the decoder to parse sps_ext_or_max_sub_layers_minus1 (see F.7.3.2.2.1). The value of this syntax element is then used to infer the value for sps_max_sub_layers_minus1 as specified in Clause F.7.4.3.2.1. The current design would led to have sps_max_sub_layers_minus1 equal to zero. This is fine for streams captured by devices such as Apple iPhones since the value for sps_ext_or_max_sub_layers_minus1 is set to 7 and the spec in this case mandates that sps_max_sub_layers_minus1 assumes the same value transmitted in the SPS. However, different encoders might be setting sps_ext_or_max_sub_layers_minus1 to a different value, thus also sps_max_sub_layers_minus1 will be different from the value parsed in the VPS. The fix proposed in av_parser.c::gf_hevc_read_sps_bs_internal addresses some discrepancy between the implementation and the spec.
- Support is added to parse and embed in a .mov file the three dimension reference displays info SEI message from Clause G.14.2.3.3. Such a supplemental information is required to play Multiview content on devices such as the Apple Vision Pro. Having MP4box embedding such information would allow users to wrap Annex B HEVC streams into .mov containers.

### DIFF
--- a/include/gpac/internal/media_dev.h
+++ b/include/gpac/internal/media_dev.h
@@ -572,6 +572,7 @@ typedef struct _hevc_state
 	u8 clli_data[4];
 	u8 mdcv_data[24];
 	u8 clli_valid, mdcv_valid;
+	u8 has_3d_ref_disp_info;
 } HEVCState;
 
 typedef struct hevc_combine{
--- a/src/filters/reframe_nalu.c
+++ b/src/filters/reframe_nalu.c
@@ -155,7 +155,7 @@ typedef struct
 	u32 nal_store_size, nal_store_alloc;
 
 	//list of param sets found
-	GF_List *sps, *pps, *vps, *sps_ext, *pps_svc, *vvc_aps_pre, *vvc_dci, *vvc_opi;
+	GF_List *sps, *pps, *vps, *sps_ext, *pps_svc, *vvc_aps_pre, *vvc_dci, *vvc_opi, *sei_prefix;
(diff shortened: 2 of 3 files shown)


jeanlf:
Thanks, this looks good to me - would you have a small public test bitstream exercising the two fixes we could use for the test suite ?

matteovn:
Sure thing Jean. I can provide a small bitstream by the end of tomorrow.  The source material will be an internal video we use in our tests. One thing is that the source will have our logo (Visionular) in it. Is that okay with you? Thanks

jeanlf:
if that's OK with you that's fine with us :)

matteovn:
No problem, the bitstream attached to this comment has sps_ext_or_max_sub_layers_minus1 = 3 and contains the 3D reference displays SEI message. The particular value sps_ext_or_max_sub_layers_minus1 would make the current master to throw an error when parsing the SPS for Layer 1 (i.e. the right view). The patch provide would instead make the code to carry as expected.
[hockeymatch_test.zip](https://github.com/user-attachments/files/16085377/hockeymatch_test.zip)

matteovn:
Many thanks Jean for your kind support :). Have a nice day

jeanlf:
Thanks, merging and adding test

PR #2865, opened 2024-06-07
Support packaging of generic ID3 tags into ISOBMFF coming from TS streams.

jadarve:
Currently, the ID3 demuxing and packaging implementation only supports Nielsen-specific tags with a fixed data size and scheme URI.

This URI improves upon those changes by supporting generic tags of any size. The TS demuxer detects whether the incoming ID3 tag corresponds to a Nielsen ID3 tag, and sets its corresponding scheme URI for later use. Otherwise, it uses `https://aomedia.org/emsg/ID3` for both scheme and value URI.

### DIFF
--- a/src/filters/dmx_m2ts.c
+++ b/src/filters/dmx_m2ts.c
@@ -33,6 +33,19 @@
 #include <gpac/thread.h>
 #include <gpac/internal/media_dev.h>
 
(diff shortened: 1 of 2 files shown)


rbouqueau:
Thanks. I've created a new branch, see new PR here: https://github.com/gpac/gpac/pull/2945

PR #2879, opened 2024-06-15
Force I-frame on SCTE-35 breaks

DenizUgur:
Added a new option for `scte35dec` that would copy the incoming packets to its output. During this operation, if a SCTE-35 splice point is encountered, that packet is marked to be encoded as an I-frame.

I wanted to use `GF_PROP_PCK_CUE_START` but that forces the encoder to reconfigure and during my tests `libx264` entered into a reconfiguration loop.

- [ ] Test with more content 
- [x] Decide on the option name
- [x] Find out how to work with `GF_PROP_PCK_CUE_START`
- [x] Not all events trigger IDR, need to handle that scenario
- [x] unit tests
- [ ] test suite


src/filters/dec_scte35.c
```diff
@@ -550,6 +578,7 @@ static const GF_FilterCapability SCTE35DecCaps[] =
 #define OFFS(_n)	#_n, offsetof(SCTE35DecCtx, _n)
 static const GF_FilterArgs SCTE35DecArgs[] =
 {
+	{ OFFS(pass), "pass-through mode, sets 'CueStart' at each splice point", GF_PROP_BOOL, "false", NULL, 0},
```
rbouqueau, line 581: I would rather call it 'mode' and explicit that one mode is 23001-18 and the other is video passthru adding the 'sctb' property at event time.

Opinion @jeanlf ?

src/filters/dec_scte35.c
```diff
@@ -470,6 +473,15 @@ static GF_Err scte35dec_process_emsg(SCTE35DecCtx *ctx, const GF_PropertyValue *
 	return GF_OK;
 }
 
+static Bool scte35dec_is_splice_point(SCTE35DecCtx *ctx, u64 dts)
+{
+	Event *evt = gf_list_get(ctx->ordered_events, 0);
+	if (!evt) return GF_FALSE;
+	Bool is_splice = (evt->dts + evt->emib->presentation_time_delta == dts);
```
rbouqueau, line 480: This compares DTS and PTS. Luckily pts=dts for most IDRs but not all (cf PTS-offset put by Elemental in their HLS streams).

Also not all events trigger an IDR, we need investigate that.

DenizUgur: This comparison didn't feel natural to me as well. Honestly I set this to CTS initially but within this filter all comparisons were made with DTS so I kept it same.

https://github.com/gpac/gpac/blob/21284f79f61119bbc8dbe7669a832b71b754a90a/src/filters/dec_scte35.c#L185

rbouqueau: dts <= pts (because you need to decode before being able to present). That's why it seems an ok estimation when deciding to flush the box. But it is not when pushing an IDR on an exact frame.

We should certainly consider getting the cts/pts as well and add some unit tests.

DenizUgur: Gotcha, I'll revert to my initial version with CTS then

DenizUgur: Also this presentation time offset is an offset from the DTS of this packet right? Or is it it's CTS? I may have missed this information on the specs

DenizUgur: I've modified the code so that only the following conditions instruct an IDR:

- `splice_insert` and `splice_event_cancel_indicator == 0`
- `time_signal` and `segmentation_type_id == 0x10`
- `time_signal` and `segmentation_type_id == 0x34`

DenizUgur:
@jeanlf any idea why `GF_PROP_PCK_CUE_START` puts `ff_enc` into reconfiguration loop? From what I can tell, the encoder is trying to consume the same packet repeatedly and that causes this loop. There might be other stuff as well.

DenizUgur:
Last commit fixed the encoder reconfiguration issue. It can be cherry-picked from this PR if it's satisfactory.

src/filters/ff_enc.c
```diff
@@ -496,6 +497,16 @@ static GF_Err ffenc_process_video(GF_Filter *filter, struct _gf_ffenc_ctx *ctx)
...
+	//don't repeat IDR sync if we already forced one
+	if (force_intra==2) {
+		if (ctx->prop_induced_reconfig) {
+			force_intra = 1;
+			ctx->prop_induced_reconfig = GF_FALSE;
+		} else {
+			ctx->prop_induced_reconfig = GF_TRUE;
+		}
+	}
+
```
jeanlf, line 509: nice catch - can you do a separate PR for this one ?

DenizUgur: Done in #2888

src/filters/dec_scte35.c
DenizUgur: like `mode=23001-18` and `mode=passthru`?

rbouqueau (approved):
LGTM

cconcolato:
@DenizUgur can you add a command line example?

DenizUgur:
The following should be enough. Pass `-graph` option to visualize if everything is connecting sequentially. If not, add `@` between filters that don't connect as desired.

```
gpac -i video.ts dec_scte35:mode=passthrough c=avc -o output.ts
```

PR #2880, opened 2024-06-17
fix imp -d relative to current directory

### LINKED ISSUE #2856: Import directory in WASM demo is broken
`imp -d` in the wasm cmd line throws exception

nlsdvl:
fixes #2856 - Import directory in WASM demo is broken

In the [WASM console](https://wasm.gpac.io), when importing a directory using the `imp -d [NAME]` command, the imported files were always created relative to the root directory, instead of being created relative to the current directory. 

This PR implements the current directory as the default import destination when NAME  is undefined, otherwise it checks if NAME is an absolute or relative path and behaves accordingly.

### DIFF
--- a/share/emscripten/gpac.html
+++ b/share/emscripten/gpac.html
@@ -250,7 +250,6 @@
           onkeypress="
           if (event.keyCode == 13) {
             if (!is_custfs) {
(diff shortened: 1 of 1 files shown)


jeanlf:
Thanks !

PR #2882, opened 2024-06-17
fix: double-free, bad-free, and heap-use-after-free issues

touatily:
fix `double-free`, `bad-free`, and `heap-use-after-free` issues

autofuzz bug `hh4LjpASSgbe5w`; `m9ncfi-rO16uCA`; `2du1EHRvDJl0Xw`; `EO-JJdpd4pGtwA`


jeanlf:
thanks !

PR #2888, opened 2024-06-20
Preventing FFmpeg encoder from unnecessary reconfiguring

DenizUgur:
As part of #2879 I've discovered that if `GF_PROP_PCK_CUE_START` was present in a packet, encoder would go in a reconfiguration loop


jeanlf:
Thanks!

PR #2893, opened 2024-06-27
export `gf_props_parse_value`

DenizUgur:
`filters.h` already includes the definition. Having this function available would be great for applications using libgpac and pass PID properties via string.


jeanlf:
you also need to export it in gpac/src/export.cpp

DenizUgur:
Done

PR #2894, opened 2024-06-27
Add `availabilityTimeComplete` to SegmentTemplate

DenizUgur:
As indicated by [CR-Low-Latency-Live-r8.pdf](https://dashif.org/docs/CR-Low-Latency-Live-r8.pdf) 9.X.4.5.5 `availabilityTimeComplete` must be present and be set to `false`


ab2022:
I tested this patch and it fixed the problem. `availabilityTimeComplete` is now in the mpd when "asto" option given

jeanlf:
thanks !

PR #2903, opened 2024-07-02
Add prft box outside dashing

DenizUgur:
`prft` box is useful outside dashing as well. It allows precise timing information for custom protocols.


DenizUgur:
Done. It is disabled if not in fragmented mode

jeanlf:
thanks!

PR #2908, opened 2024-07-06
uncompressed: add icbr and cmpC box dump

bradh:
From 23001-17 draft AMD2.

Also includes some minor typo fixes I noted while reading. I'll split those out if you'd prefer.

### DIFF
--- a/src/isomedia/box_dump.c
+++ b/src/isomedia/box_dump.c
@@ -1840,6 +1840,25 @@ static GF_Err dump_itai(GF_UnknownBox *u, FILE * trace)
 	return GF_OK;
 }
 
+static GF_Err dump_cmpc(GF_UnknownBox *u, FILE * trace)
+{
+	u32 val;
+	GF_BitStream *bs = gf_bs_new(u->data, u->dataSize, GF_BITSTREAM_READ);
+	gf_isom_box_dump_start((GF_Box *)u, "CompressionConfigurationBox", trace);
+
+	//full box
+	get_and_print("version", 8)
+	get_and_print("flags", 24)
+	get_4cc_and_print("compression_type", 32)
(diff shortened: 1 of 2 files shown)


bradh:
Some sample files.
[rgb_generic_compressed_samples.zip](https://github.com/user-attachments/files/16117548/rgb_generic_compressed_samples.zip)

jeanlf:
Could you split the typo fixes in a seperate PR ? We'll wait for AMD2 to be frozen before adding support for these boxes.

bradh:
Typos now in #2913. Will convert this to a draft and update as required.

bradh:
I think AMD 2 has progressed enough to add this.

Have updated the implementation to match the DAM, except for `cmpC` where I omitted the `must_decompress_individual_units`.

I am unsure what is the best way to handle the implied offset (i.e. where `unit_offset_nbbits` is 0). In this implementation, I have tracked the implied offsets, and included them as if they were present. However it might also be possible to just not show the offset. Review feedback appreciated.

jeanlf:
> Have updated the implementation to match the DAM, except for cmpC where I omitted the must_decompress_individual_units.

I think we should dump it as it is still present in the AMD2 text

> However it might also be possible to just not show the offset. 

I'd rather go for this, and only dump offsets when present in the box.

Can you update the PR to reflect these?

bradh:
> > Have updated the implementation to match the DAM, except for cmpC where I omitted the must_decompress_individual_units.
> 
> I think we should dump it as it is still present in the AMD2 text

I have made this change, although I could see it both ways.
 
> I'd rather go for this, and only dump offsets when present in the box.

Fixed.

jeanlf:
Thanks !

PR #2932, opened 2024-08-01
Segment Sequence Representations for DASH

DenizUgur:
So far, the manifest seems to be generated without a problem. The new option is `dashll=sf`.

Shaka player [hyperlink](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;lowLatencyMode=true;autoLowLatencyMode=true;forceHTTP=true;parsePrftBox=true;streaming.liveSync.enabled=true;infiniteLiveStreamDuration=true;uilang=en-US;asset=http://localhost:8080/test.mpd;panel=CUSTOM%20CONTENT;build=uncompiled)

## TODO
- [x] If segments don't have consistent parts, ~we have to warn that the input frame rate is not correct~ new `<S />` elements must be generated
- [ ] Implement playback with gpac

## Testing with
```
gpac avgen:lock:v c=libx264:fintra=1 reframer:rt=on -o http://localhost:8080/test.mpd:gpac:profile=dashif.ll:asto=1.7:dmode=dynauto:stl:segdur=2:cdur=0.1:ntp=yes:cmf2:rdirs=/tmp:dashll=sf
```

### DIFF
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -7304,12 +7304,13 @@ static void dasher_insert_timeline_entry(GF_DasherCtx *ctx, GF_DashStream *ds)
 	Bool is_first = GF_FALSE;
 	Bool seg_align = GF_FALSE;
 	GF_MPD_SegmentTimeline *tl=NULL;
+	GF_DASH_SegmentContext *sctx;
 
 	//we only store segment timeline for the main component in the representation
 	if (ds->muxed_base) return;
 
 	if (ds->rep && ds->rep->state_seg_list) {
-		GF_DASH_SegmentContext *sctx = gf_list_last(ds->rep->state_seg_list);
+		sctx = gf_list_last(ds->rep->state_seg_list);
 		if (sctx)
 			sctx->dur = ds->first_cts_in_next_seg - ds->first_cts_in_seg;
 	}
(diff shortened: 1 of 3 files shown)


src/filters/dasher.c
```diff
@@ -7458,7 +7468,9 @@ static void dasher_insert_timeline_entry(GF_DasherCtx *ctx, GF_DashStream *ds)
 	if (!s) return;
 	s->start_time = ds->seg_start_time + pto;
 	s->duration = (u32) duration;
+	if (ctx->dashll) s->nb_parts = sctx->nb_frags + 1;
```
DenizUgur, line 7471: I have to add `sctx->nb_frags + 1` because there's always one more fragment left to be processed. Could be a bug.

src/filters/dasher.c
```diff
@@ -10775,6 +10784,10 @@ static const GF_FilterArgs DasherArgs[] =
 		"- br: use LL-HLS with byte-range for segment parts, pointing to full segment (DASH-LL compatible)\n"
 		"- sf: use separate files for segment parts (post-fixed .1, .2 etc.)\n"
 		"- brsf: generate two sets of manifest, one for byte-range and one for files (`_IF` added before extension of manifest)", GF_PROP_UINT, "off", "off|br|sf|brsf", GF_FS_ARG_HINT_EXPERT},
+	{ OFFS(dashll), "DASH Low Latency type\n"
+		" - cte: use Chunked Transfer Encoding for segments\n"
+		" - sf: use separate files for segment parts (post-fixed .1, .2 etc.)\n",
```
rbouqueau, line 10798: I think we should use the word "Segment Sequence" somewhere to make it searchable for users

Issue #2698, opened 2023-12-01
 Infinite loop in av1_uvlc media_tools/av_parsers.c:1664


hu1y40:
1.Version
MP4Box - GPAC version 2.3-DEV-rev636-gfbd7e13aa-master
(c) 2000-2023 Telecom Paris distributed under LGPL v2.1+ - https://gpac.io

Please cite our work in your research:
	GPAC Filters: https://doi.org/10.1145/3339825.3394929
	GPAC: https://doi.org/10.1145/1291233.1291452

GPAC Configuration: --static-bin --enable-sanitizer
Features: GPAC_CONFIG_LINUX GPAC_64_BITS GPAC_HAS_IPV6 GPAC_HAS_SOCK_UN GPAC_MINIMAL_ODF GPAC_HAS_QJS GPAC_HAS_LINUX_DVB  GPAC_DISABLE_3D

2.Program Output
[iso file] Parsing a top-level box at position 0
[iso file] Read Box type ftyp size 32 start 0
[iso file] Parsing a top-level box at position 32
[iso file] Read Box type mv@d size 108 start 32
[iso file] Unknown top-level box type mv@d
[iso file] Parsing a top-level box at position 140
[iso file] Read Box type av1C size 363 start 140
[AV1] parsed AV1 OBU type=7 size=99 at position 152.
[AV1] AV1 unexpected OBU type=152 size=0 found at position 140056310360544. Forwarding.
[AV1] parsed AV1 OBU type=1 size=12 at position 251.
[AV1] parsed AV1 OBU type=1 size=12 at position 263.
[AV1] parsed AV1 OBU type=1 size=12 at position 275.
[AV1] parsed AV1 OBU type=1 size=12 at position 287.
[AV1] parsed AV1 OBU type=2 size=25 at position 299.
[AV1] AV1 unexpected OBU type=299 size=0 found at position 140056310360544. Forwarding.
[AV1] parsed AV1 OBU type=2 size=52 at position 324.
[AV1] AV1 unexpected OBU type=324 size=0 found at position 140056310360544. Forwarding.
[AV1] parsed AV1 OBU type=1 size=12 at position 376.
[AV1] parsed AV1 OBU type=1 size=12 at position 388.
[AV1] parsed AV1 OBU type=1 size=12 at position 400.
[AV1] parsed AV1 OBU type=1 size=12 at position 412.
[AV1] parsed AV1 OBU type=1 size=12 at position 424.
[...]

PR #2937, opened 2024-08-08
backport the security patch of CVE-2023-50120

Crispy-fried-chicken:
Here is a vulnerability which is mentioned in https://github.com/gpac/gpac/issues/2698 and fixed in the master branch(https://github.com/gpac/gpac/commit/b655955b840ccd7c7198bb15375aa510e76208eb) but is not fixed in the branch of buildbot-mabr_misc,  maybe it should be backported?


aureliendavid:
This was a development branch that was merged into master in https://github.com/gpac/gpac/pull/2910 

So the branch is now stale, it will either be deleted or rebased to master if further developments are needed. 

So no need to backport stuff into that branch unless i'm missing something.

PR #2940, opened 2024-08-21
fix possible overflow in dash_client 

soheibthriber:
I changed the static declaration of szTemplate char pointer to a dynamic declaration.

### DIFF
--- a/src/media_tools/dash_client.c
+++ b/src/media_tools/dash_client.c
@@ -847,7 +847,9 @@ setup_multicast_clock:
 				len = (u32) strlen(seg_url);
 				if (!strncmp(val, seg_url, len)) {
 					u64 number=0;
-					char szTemplate[100];
+					int template_len = strlen(seg_url) + 20; // Allocate extra space for "%"
+					char *szTemplate;
+					GF_SAFE_ALLOC_N(szTemplate, template_len, char);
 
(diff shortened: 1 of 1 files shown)


src/media_tools/dash_client.c
```diff
@@ -847,7 +847,9 @@ static void gf_dash_group_timeline_setup_single(GF_MPD *mpd, GF_DASH_Group *grou
 				len = (u32) strlen(seg_url);
 				if (!strncmp(val, seg_url, len)) {
 					u64 number=0;
-					char szTemplate[100];
+					int template_len = strlen(seg_url) + 20; // Allocate extra space for "%"
```
aureliendavid, line 850: Hi, 

I'm not sure about this (haven't tested) so I might be mistaken but I'm wondering if the length you want here is the original length of seg_url (+ 20). 

But on line 846 we're cutting `seg_url` at the `start` pointer. So the `strlen(seg_url)` here might be too small? Maybe we should save the `strlen(seg_url)` before the start[0} = 0 to allocate the correct length here. 

I may be incorrect though.

jeanlf:
Could you post a command line excising this overflow ?

aureliendavid:
Thanks for the data. 

The bug doesn't trigger with the pcap because it doesn't have the long base url but I was able to reproduce it with a custom mpd with a long url. 

I was also able to trigger the caveat I pointed above so I added the fix I suggested here https://github.com/gpac/gpac/commit/6b4eea93748b44f89df41e0c6e7187ee970be2b5

PR #2941, opened 2024-08-21
Disabling Chunked Transfer Encoding

### LINKED ISSUE #2939: Disabling Chunked Transfer Encoding globally
I wonder if it's possible to disable CTE altogether. In case the origin server is using a non-compliant HTTP server and gpac pushes to this server via `hmode=push`, we might require the option to disable CTE. Similar to how we disable HTTP/2 upgrades via `-no-h2c`. I had an idea to make this a reality by skipping connection creation when httpout opens the input and only process the connection when closing the input.

Any ideas to make this feature a reality? Is there a better way to achieve this?

DenizUgur:
I had some success in disabling the `Transfer-Encoding: chunked` header for upload requests. @jeanlf I'd really appreciate your input on this.

Adding support for HTTP/2 should be straightforward but I'll look into it later.

## Usage

Add the `-no-cte` global option to disable chunked transfer altogether.

## TODO
- [ ] Test with HTTP/2
- [ ] Defer connection setup until data is ready
- [x] Test with DELETE requests. Should be okay because we check if the request has body but verify again

closes #2939

### DIFF
--- a/src/utils/downloader.c
+++ b/src/utils/downloader.c
@@ -5485,7 +5485,7 @@ static GF_Err http_send_headers(GF_DownloadSession *sess, char * sHTTP) {
 	} else if (!strcmp(req_name, "HEAD")) sess->http_read_type = HEAD;
 	else sess->http_read_type = OTHER;
 
-	if (!strcmp(req_name, "PUT") || !strcmp(req_name, "POST"))
+	if ((!strcmp(req_name, "PUT") || !strcmp(req_name, "POST")) && !gf_opts_get_bool("core", "no-cte"))
 		sess->put_state = 1;
 
 	//note that url is not used for CURL, already setup together with proxy
--- a/src/utils/downloader.c
+++ b/src/utils/downloader.c
@@ -264,6 +264,10 @@ struct __gf_download_session
 	//0: not PUT/POST, 1: waiting for body to be completed, 2: body done
 	u32 put_state;
 
+	void* holdback_data;
+	u32 holdback_data_size;
+	Bool holdback_data_complete;
+
 	u64 last_cap_rate_time;
 	u64 last_cap_rate_bytes;
 	u32 last_cap_rate_bytes_per_sec;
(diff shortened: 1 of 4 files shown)


src/utils/downloader.c
```diff
@@ -7774,8 +7821,20 @@ GF_Err gf_dm_sess_send(GF_DownloadSession *sess, u8 *data, u32 size)
...
 #endif
 
+	if (sess->should_holdback_data && !sess->holdback_data_complete) {
+		if (!data || !size)
+			sess->holdback_data_complete = GF_TRUE;
+
+		if (!sess->holdback_data_complete) {
+			sess->holdback_data_size += size;
+			sess->holdback_data = gf_realloc(sess->holdback_data, sess->holdback_data_size);
+			memcpy(sess->holdback_data + sess->holdback_data_size - size, data, size);
```
roman380, line 7831: I had to change here:

```
			memcpy((u8*) sess->holdback_data + sess->holdback_data_size - size, data, size);
```

DenizUgur: I assume you were getting segfaults otherwise?

roman380: No, I just could not build it

roman380: > 1>C:\Project\castlabs.com\gpac.io\gpac_public\src\utils\downloader.c(7831,31): error C2036: 'void *': unknown size
1>C:\Project\castlabs.com\gpac.io\gpac_public\src\utils\downloader.c(7831,58): error C2036: 'void *': unknown size

DenizUgur: My compiler didn't complain about this. I'll push this change though

jeanlf:
Given that only httpout uses CTE upload, I think it would be simpler/better to prevent CTE in the httpout filter directly. This would take care of payload reconstruction by forcing input pid framing? What do you think ?

DenizUgur:
My initial attempt was within httpout but it seemed too complicated to do it there. If you'd provide some guidance, I'd be happy to modify it.

jeanlf:
I added `cte` (bool, default=true) option in httpout to control chunk transfer mode without changing the downloader. Option works for upload (hmode=push) and for regular server as well. Note that for the time being, we cannot prevent CTE when a GET is made on a resource currently being uploaded by PUT/POST.

I'm closing this, please open an issue if needed.

PR #2955, opened 2024-09-06
Treat SAPs as Cue Points

DenizUgur:
Opened in place of #2950

By placing a reframer with `:sapcue=3`, we signal that the dasher may use this information to define segment boundaries. The value given to `sapcue` will determine which SAPs get treated as cues. SAPs smaller than or equal to a given value will signal cues. However, for that to happen dasher must be configured to use in-band cues via `:cues=inband`

### DIFF
--- a/src/filters/reframer.c
+++ b/src/filters/reframer.c
@@ -152,6 +152,7 @@ typedef struct
 	GF_PropStringList props;
 	Bool copy, rmseek;
 	u32 cues;
+	Bool sapcue;
 
 	//internal
 	Bool filter_sap1;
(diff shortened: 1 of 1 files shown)


src/filters/reframer.c
```diff
@@ -2774,6 +2787,7 @@ static const GF_FilterArgs ReframerArgs[] =
 	"- no: do no filter frames based on cue info\n"
 	"- segs: only forward frames marked as segment start\n"
 	"- frags: only forward frames marked as fragment start", GF_PROP_UINT, "no", "no|segs|frags", GF_FS_ARG_HINT_EXPERT|GF_FS_ARG_UPDATE},
+	{ OFFS(sapcue), "treat SAPs as cue points", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_EXPERT },
```
jeanlf, line 2790: this should not be a boolean but rather a SAP value: 0 (default) disables, otherwise only set cue if sap>0 and sap<=sapcue

jeanlf (changes requested):

jeanlf:
Thanks!

PR #2956, opened 2024-09-06
dasher shouldn't anticipate segment boundary on video

rbouqueau:
There are 3 distincts commits in this PR that could be cherry-picked if the PR is not acceptable.

### DIFF
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -7268,7 +7268,7 @@ static GF_Err dasher_setup_period(GF_Filter *filter, GF_DasherCtx *ctx, GF_DashS
 		//in sbound=0 mode, if stream has sync and non-sync and uses skip samples, allow spliting
 		//slightly before - typically needed for audio with sync points (usac, mpegh) where the segment duration is set
 		//to the intra interval, we need to take into account the skip samples
-		if (!ctx->sbound && !ds->cues && (ds->sync_points_type==DASHER_SYNC_PRESENT)
+		if (!ctx->sbound && !ds->cues && (ds->sync_points_type==DASHER_SYNC_PRESENT) && ds->stream_type==GF_STREAM_AUDIO
 			&& (ds->pts_minus_cts<0) && (ds->next_seg_start > (u32) -ds->pts_minus_cts)
 		) {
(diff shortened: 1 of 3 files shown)


aureliendavid:
Hi, 

one small thing: when branch names contains underscores (and probably other characters) it trips the debian packaging script and the buildbot process fails for the linux workers 

> dh_strip: warning:     debian/changelog(l1): version '2.5-DEV-rev506-g6da37dcc-buildbot-reframer_sap' is invalid: version number contains illegal character '_'

we should probably add a fix to this but in the meantime for buildbot-* branches it's better to stick to alphanum + dashes


other than that, if we look at the tests that did run (https://tests.gpac.io/revision/2.5-DEV-rev506-g6da37dcc-buildbot-reframer_sap and check "Hide lines with any OK") it looks like the tests for: 

* add-subsamples-HEVC
* shvc
* svc

result in a different hashes than on master. 

It might be intended that the correct result for these tests has changed with this PR, in which case a PR in the testsuite repo with the new correct hashes would be appreciated. 

Or it might be a bug in the PR that can be fixed here.

rbouqueau:
@aureliendavid Thank you. Indeed the hashes changed.

Waiting for @jeanlf 's feedback.

jeanlf:
OK with the first two commits, but I'm not sure what you're trying to fix on the dasher. Would you have a sample faulty stream ?

rbouqueau:
```
$ gpac -i https://dash-large-files.akamaized.net/dvb-ibc-2024/13%20-%205G%20Broadcast%20Booth%20Version.mp4 c=avc:fintra=2:FID=V1 -o dash.mpd:SID=V1:profile=dashavc264.live:segdur=2
No filter chain found for PID A2 in filter mp4dmx to any loaded filters - NOT CONNECTED
[Dasher] WARNING! Max SAP type 3 detected - switching to FULL profile
```

We detect SAP type 3 because GPAC switches too early (ie not on the `fintra` SAP1).

jeanlf:
OK I think this is correct, the patch is indeed mostly for audio but should not apply to video. Maybe we'll need a way to configure this through an option in case we encounter audio encodings that include the initial priming duration in the first SAP period.

PR #2962, opened 2024-09-16
Write empty WebVTT file if no cue is present

DenizUgur:
Add the ability to export the WebVTT track when no cues are present in the track.

```
MP4Box -raw <vtt_track_id> --noempty=0 test.mp4
gpac -i test.mp4 @#text -o test.vtt:!noempty
```

### DIFF
--- a/src/filters/write_vtt.c
+++ b/src/filters/write_vtt.c
@@ -55,6 +55,7 @@ typedef struct
 
 	GF_FilterPacket *src_pck;
 	Bool dash_mode;
+	Bool first;
 	u32 seg_pck_in, seg_pck_out;
 } GF_WebVTTMxCtx;
 
@@ -80,6 +81,7 @@ GF_Err vttmx_configure_pid(GF_Filter *filter, GF_FilterPid *pid, Bool is_remove)
 	if (!p) return GF_NOT_SUPPORTED;
 	ctx->codecid = p->value.uint;
 	ctx->ipid = pid;
+	ctx->first = GF_TRUE;
 
 
 	if (!ctx->opid) {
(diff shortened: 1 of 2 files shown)


src/filters/write_vtt.c
```diff
@@ -235,6 +237,14 @@ GF_Err vttmx_process(GF_Filter *filter)
 	if (!pck) {
 		if (gf_filter_pid_is_eos(ctx->ipid)) {
 			gf_filter_pid_set_eos(ctx->opid);
```
jeanlf, line 239: set_eos should be after your injected packet

src/filters/write_vtt.c
```diff
@@ -247,6 +257,7 @@ GF_Err vttmx_process(GF_Filter *filter)
 		return GF_OK;
 	}
 	data = (char *) gf_filter_pck_get_data(pck, &pck_size);
+	ctx->first = GF_FALSE;
```
jeanlf, line 260: we should have on option on vttmx to enable this feature, for cases where we want to dump exactly what is in the stream.

I'd suggest replacing `first` by `noempty` and exporting it as an option in WebVTTMxArgs

jeanlf (changes requested):

src/filters/write_vtt.c
```diff
@@ -236,15 +234,14 @@ GF_Err vttmx_process(GF_Filter *filter)
 	pck = gf_filter_pid_get_packet(ctx->ipid);
 	if (!pck) {
 		if (gf_filter_pid_is_eos(ctx->ipid)) {
-			gf_filter_pid_set_eos(ctx->opid);
-			if (ctx->first) {
+			if (!ctx->noempty) {
```
jeanlf, line 237: if :!noempty is set, this will always inject an empty VTT on EOS even if some VTT cues have been sent, which is not desired. You need to combine this with a counter of packets sent

jeanlf (changes requested):

src/filters/write_vtt.c
DenizUgur: True, I guess when I removed `ctx->first` I forgot about that

jeanlf:
thanks

cconcolato:
@DenizUgur why do we need a parameter? With your patch, no file would produced at all in the default case. That is unexpected as MP4Box always produces something. Is there a use case where we don't want to generate files just with the WebVTT signature? if not, I would simply remove the flag or make it false by default.

DenizUgur:
My initial version was to make this work without any option. @jeanlf could pitch it, maybe he didn't want to alter how MP4Box is working right now.

cconcolato:
@DenizUgur / @jeanlf can you follow-up?

DenizUgur:
I would recommend that we keep this behavior for tests but disable `noempty` by default.

DenizUgur:
In the mean time would using a [configuration file](https://wiki.gpac.io/Filters/core_config/) help?

rbouqueau:
@jeanlf noempty should be the default, shouldn't it?

PR #2965, opened 2024-09-17
fix a bunch of use-of-uninitialized-memory errors

aureliendavid:
Hi all,

For clarity and record keeping, I'm putting all of these in a single PR, with each bug in its own commit. 

## Credits

Thanks and credits to @skorpion98 and @Heinzeen from Sapienza University of Rome for the detection and responsible disclosure of these issues. 

## Security and functional concerns

I don't think any of these poses great security concerns since the path to exploitation seems really difficult. 

However we've had functional seemingly-random bugs caused by uninitialized variables before, so these reports are valuable to remove some sources of undefined behavior. 

## Good practices for developers

Sorry in advance if the recommendations in this section seem trivial or obvious, but the existence of these issues show that we need to be more vigilant on a few points: 

### Init local structures

When using local variables in a block, remember to give them an initial value (0/-1 for ints, null for pointers, etc.). 

For structs, it's important to remember that without OOP-style constructors, fields inside structs are not initialized. 

For struct pointers on the heap we usually use a constructor function (like `gf_bs_new()`) or the `GF_SAFEALLOC()` macro. 

For local structs on the stack we need to manually init the fields to a sensible value. We can also memset() the whole struct to 0 to just ensure initialization. e.g.:

```c
GF_Fraction frac; 
frac.num=0; frac.den=1;

FLACHeader hdr;
memset(&hdr, 0, sizeof(FLACHeader));
```

### check buffer functions

One use case that happens a lot in gpac is: 

1. we declare a pretty big buffer
2. we get the functional size from somewhere (bitstream read for example)
3. we read/write size bytes of data into our buffer
4. we pass this buffer and size to another function/struct

It's important to have checks around point 2 since it usually comes from user supplied data. 

But for our subject matter here, it's on point 3 that we need to be more careful. 

Functions like `fread()`, `gf_bs_read_data()`, and many buffer handling functions will return the number of bytes actually read (or written). 

We need to think about checking this return value against the hoped size we passed, so that we can know which part of the buffer is actually initialized. 

Again these are pretty basic but it never hurts to be reminded of good practices :)

### DIFF
--- a/src/filters/inspect.c
+++ b/src/filters/inspect.c
@@ -3316,7 +3316,7 @@ static void inspect_dump_vpx(GF_InspectCtx *ctx, FILE *dump, u8 *ptr, u64 frame_
 {
 	GF_Err e;
 	Bool key_frame = GF_FALSE;
-	u32 width = 0, height = 0, renderWidth, renderHeight;
+	u32 width = 0, height = 0, renderWidth=0, renderHeight=0;
 	u32 num_frames_in_superframe = 0, superframe_index_size = 0, i = 0;
 	u32 frame_sizes[VP9_MAX_FRAMES_IN_SUPERFRAME];
 	gf_bs_reassign_buffer(pctx->bs, ptr, frame_size);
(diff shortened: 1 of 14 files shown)


rbouqueau:
Side note: since we're using C99 we can replace `memset()` with `FLACHeader hdr = {0};`.

PR #2968, opened 2024-09-18
add parsing and writing iamf and iacb boxes

felicialim:
IAMF is a codec-agnostic container specification for immersive spatial audio developed by the Alliance for Open Media (AOMedia).

The IAMF v1.0.0-errata spec is at https://aomediacodec.github.io/iamf/v1.0.0-errata.html.


src/isomedia/box_code_base.c
```diff
@@ -12713,6 +12714,67 @@ GF_Err dOps_box_size(GF_Box *s)
...
+
+        ptr->cfg = gf_odf_ia_cfg_read_bs_size(bs, (u32)ptr->size);
+
+        read = gf_bs_get_position(bs) - pos;
+
+        if (read < ptr->size)
+		GF_LOG(GF_LOG_WARNING, GF_LOG_CONTAINER, ("[ISOBMFF] IAConfigurationBox: read only "LLU" bytes (expected "LLU").\n", read, ptr->size));
+        if (read > ptr->size)
+		GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("[ISOBMFF] IAConfigurationBox overflow read "LLU" bytes, of box size "LLU".\n", read, ptr->size));
+
```
jeanlf, line 12747: these two tests are not needed, they are already done by the generic box parse function

jeanlf (changes requested):

jeanlf:
Do you have any public test stream we could integrate in the test suite ? Ultimately we would need a reframe_iamf to parse these, have you looked into this ?

rbouqueau:
I see a test here: https://github.com/gpac/testsuite/pull/18

src/media_tools/av_parsers.c
```diff
@@ -4385,6 +4385,133 @@ GF_Err gf_media_prores_parse_bs(GF_BitStream *bs, GF_ProResFrameInfo *prores_fra
...
+
+        if (gf_bs_is_overflow(bs)) {
+              GF_LOG(GF_LOG_WARNING, GF_LOG_CODING, ("[IAMF] OBU parsing consumed too many bytes.\n"));
+              e = GF_NON_COMPLIANT_BITSTREAM;
+        }
+        gf_bs_seek(bs, pos + *obu_size);
+        return e;
+}
+
+GF_EXPORT
```
rbouqueau, line 4453: I think this function could be static (and removed from the header). Any reason to export it?

src/isomedia/box_code_base.c
```diff
@@ -24,6 +24,7 @@
  */
 
 #include <gpac/internal/isomedia_dev.h>
+#include <gpac/internal/media_dev.h>
```
rbouqueau, line 27: @jeanlf Should we need this? or should some code from this PR be moved?

src/isomedia/isom_write.c
```diff
@@ -4434,6 +4435,11 @@ GF_Err gf_isom_clone_track(GF_ISOFile *orig_file, u32 orig_track, GF_ISOFile *de
 	if (dest_file->moov->mvhd->nextTrackID <= new_tk->Header->trackID)
 		dest_file->moov->mvhd->nextTrackID = new_tk->Header->trackID+1;
 
+        /*if it contains IAMF, add the iamf brand to ftyp*/
+        if (gf_isom_get_media_subtype(dest_file, new_tk->Header->trackID, 1) == GF_ISOM_BOX_TYPE_IAMF) {
```
rbouqueau, line 4439: This compares a subtype and a box type.

src/isomedia/box_code_base.c
felicialim: Done

src/isomedia/isom_write.c
felicialim: Thanks, I missed this. Looks like I could just iterate through the SampleDescription's child boxes and check their type.

src/media_tools/av_parsers.c
felicialim: you're right, changed to be static and removed from header

felicialim:
> Do you have any public test stream we could integrate in the test suite ? Ultimately we would need a reframe_iamf to parse these, have you looked into this ?

In addition to https://github.com/gpac/testsuite/pull/18, there are more at https://github.com/AOMediaCodec/libiamf/tree/main/tests if needed. I haven't looked at reframe_iamf yet.

src/isomedia/isom_write.c
jeanlf: No, you just need to declare a subtype for IAMF in isomedia.h (with same 4CC as the internal box type)

felicialim: Done

src/isomedia/box_code_base.c
jeanlf: no we shouldn't this comes from gf_av1_leb128_size I guess

src/isomedia/box_code_base.c
```diff
@@ -12713,6 +12714,62 @@ GF_Err dOps_box_size(GF_Box *s)
...
+GF_Err iacb_box_size(GF_Box *s)
+{
+        GF_IAConfigurationBox *ptr = (GF_IAConfigurationBox *)s;
+        if (!ptr->cfg) {
+            ptr->size = 0;
+            return GF_BAD_PARAM;
+        }
+        ptr->size += 1;  // configurationVersion
+        ptr->size += gf_av1_leb128_size(ptr->cfg->configOBUs_size);
+        ptr->size += ptr->cfg->configOBUs_size;
```
jeanlf, line 12769: could you add a helper function to get the size of the config and not include gpac/internal/media_dev here ?

rbouqueau:
FYI just saw FFmpeg 7.1 had support for it. We may want to ensure we interface well with them: https://jbkempf.com/blog/2024/ffmpeg-7.1.0/.

src/odf/descriptors.c
```diff
@@ -1931,3 +1931,123 @@ GF_Err gf_odf_opus_cfg_write(GF_OpusConfig *cfg, u8 **data, u32 *size)
...
+
+        gf_bs_write_u8(bs, cfg->configurationVersion);
+        gf_av1_leb128_write(bs, cfg->configOBUs_size);
+        for (int i = 0; i < gf_list_count(cfg->configOBUs); ++i) {
+                GF_IamfObu *configOBU = gf_list_get(cfg->configOBUs, i);
+                gf_bs_write_data(bs, configOBU->raw_obu_bytes, (u32)configOBU->obu_length);
+        }
+
+        return GF_OK;
+}
```
jeanlf, line 2053: Add here `u32 gf_odf_ia_cfg_size(GF_IAConfig *cfg)` (it's ok to include media_dev.h in this file

jeanlf (changes requested):

src/odf/descriptors.c
felicialim: Done. I removed the `#ifndef GPAC_DISABLE_AV_PARSERS` around the media_dev.h include but let me know it's preferred to check that inside of `gf_odf_ia_cfg_size()`.

src/isomedia/box_code_base.c
felicialim: `iacb_box_size()` now calls `gf_odf_ia_cfg_size()` instead, which removes the need for using `gf_av1_leb128_size` and having this dependency on media_dev.h

rbouqueau:
@felicialim do we agree we can close this one, that is superseeded by #3057 ?

felicialim:
> @felicialim do we agree we can close this one, that is superseeded by #3057 ?

Yes

PR #2979, opened 2024-09-28
timecode manipulations

rbouqueau:
These are random commits - most of them are already merged in master. Merging now to avoid stretching it in time (first commit was last Nov).

The code for useful timestamp manipulation is located in a new `tc_rw` filter. Actually built externally. I initially planned to write it in js. However parsing seemed too difficult as helper functions were only available in C. So switching back to C.


rbouqueau:
@jeanlf I fixed the merging issues. Good to go!

jeanlf:
thanks!

jeanlf:
@rbouqueau I'm setting the `tc` option to false by defualt in the code as it is broken: you must recompute the size of additional data injected by `tc` for proper size estimation. The current code needs fixing but is still in master (only disabled), please have a look at this asap. A good way to check this is using testsuite script/iff-grid.sh

PR #2980, opened 2024-10-01
fix for invalid file that use non-zero upid length even if the upid type is not set

cconcolato:
Example of wrong SCTE marker:
https://tools.middleman.tv/scte35-parser?hex=FC302F00000000000000FFF00506FE19206BBE0019021743554549000000057FC0000052636200016134000000009B016FB3&format=json


rbouqueau:
LGTM, thanks!

PR #2986, opened 2024-10-06
Add support for displaying all filter options using filtername.* or filtername.@ syntax

touatily:
This PR introduces a new feature that allows users to print all available options for a specific filter using the `gpac -h <filtername>.*` or `gpac -h <filtername>.@`


jeanlf:
thanks

PR #2989, opened 2024-10-07
Fix doc printing and formatting issues: Conditionally print </div>

touatily:
This PR addresses two minor issues related to documentation printing:

- The `</div>` tag is now printed only if `gen_doc==1` to avoid unnecessary HTML tags in cases where documentation generation is not required.
- Replaced spaces with tabs for consistent formatting across the codebase.


jeanlf:
thanks

PR #2994, opened 2024-10-09
Fix build issues for FFmpeg 7+

### LINKED ISSUE #2991: Build issues with ffmpeg@7
It looks like FFmpeg 7 has compatibility issues with gpac.

```bash
filters/ff_dmx.c:2219:23: fatal error: no member named 'priv_data_size' in 'struct AVInputFormat'
```

I'm on macOS with ARM chip

DenizUgur:
`avdevice_list_input_sources` was added back in libavcodec 56, we use this function in gpac only if libavcodec is 58 or greater.

Closes #2991

### DIFF
--- a/src/filters/ff_dmx.c
+++ b/src/filters/ff_dmx.c
@@ -2204,7 +2204,6 @@ char *dev_desc = NULL;
 static void ffavin_enum_devices(const char *dev_name, Bool is_audio)
 {
 	const AVInputFormat *fmt;
-	AVFormatContext *ctx;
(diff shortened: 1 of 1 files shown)


jeanlf:
shouldn't we keep the code for libavdevice<56 ?

DenizUgur:
We could, but this function only gets compiled for >58 anyway. See the `FFPROBE_DEVICES` definition

jeanlf:
Thanks!

PR #2995, opened 2024-10-09
Fixes in print_filters function

touatily:
- Enable display of `js` option in `httpout` and `jsf` filters


touatily:
The command line `gpac -h jsf:js=script.js` still doesn't work
It seems that [this if statement](https://github.com/gpac/gpac/blob/5df43d8626d8232271f4b9c31efa944185964996/applications/gpac/gpac_help.c#L2196) is causing the issue

jeanlf:
thanks, fixed in a slightly diffferent way on master

PR #2997, opened 2024-10-11
Fix inverted display of input/output filters in protocols  help

touatily:
This PR addresses an issue where the `gpac -ha protocols` and `gpac -hh protocols` commands displayed the list of input and output filters for each protocol incorrectly, with input filters shown under "out" and output filters under "in". The filter lists are now correctly categorized and displayed.


jeanlf:
Thanks!

PR #3002, opened 2024-10-17
Add warning message for blacklisted repair servers

touatily:
This PR adds a warning message to inform users when a repair server has been blacklisted because it does not support byte range requests (partial repair).


jeanlf:
thanks!

PR #3003, opened 2024-10-17
Update .gitignore to exclude Python cache files

touatily:
This PR updates the `.gitignore` file to exclude Python cache files. The following entries were added:

- `__pycache__/`: to ignore Python bytecode directories.
- `*.py[cod]`: to ignore compiled Python files.


jeanlf:
thanks!

PR #3006, opened 2024-10-18
test pls ignore

aureliendavid:
test


DenizUgur:
Should we indicate pass, warn, fail icons?

rbouqueau:
Were @DenizUgur requests covered? Should we close?

DenizUgur:
I honestly think current implementation is great, it shows the history of builbot. So I rescind my suggestion.

PR #3008, opened 2024-10-18
UTF conversion buffer sizes: fix crash in .ts -> ccdec -> .vtt conversion

rbouqueau:
```
unsigned short utf16Line = gf_malloc( sizeof(u16)  s_txt->len);
[...]
			len = gf_utf8_mbstowcs((u16*)utf16Line, 10000, (const char **) &str);
```

We have this hardcoded pattern. It seems safe because the underlying buffer sizes shouldn't allow to go beyond the allocated size. However this is weird. Does anyone remember how we ended up with this in the code?


rbouqueau:
@jeanlf Good to go!

PR #2984, opened 2024-10-03
Refactor: Change repair variable type from u32 to enum

touatily:
This PR refactors the repair variable by changing its type from `u32` to `enum`. This modification ensures better code readability and makes the logic behind the repair state more explicit, aligning the type with its intended usage.

PR #3012, opened 2024-10-21
Ensure Enum Arguments are of u32 Type

touatily:
This PR includes the following changes to improve type safety and consistency:

- fix type of `bsrw.vidfmt`: `GF_PROP_SINT` instead of `GF_PROP_UINT`.
- Declare all `enum` arguments as `u32` variables to maintain consistency and prevent type mismatches.

PS: This PR is a follow-up to the closed PR #2984

### DIFF
--- a/src/filters/mux_isom.c
+++ b/src/filters/mux_isom.c
@@ -8464,7 +8464,7 @@ static const GF_FilterArgs MP4MuxArgs[] =
 			"- v1: use v1 signaling, ISOBMFF style (will mux raw PCM as ISOBMFF style)\n"
 			"- v1qt: use v1 signaling, QTFF style\n"
 			"- v2qt: use v2 signaling, QTFF style (lpcm entry type)"
-		, GF_PROP_UINT, "v0", "v0|v0s|v1|v1qt|v2qt", 0},
+		, GF_PROP_UINT, "v0", "|v0|v0s|v1|v1qt|v2qt", 0},
 	{ OFFS(ssix), "create `ssix` box when `sidx` box is present, level 1 mapping I-frames byte ranges, level 0xFF mapping the rest", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_ADVANCED},
 	{ OFFS(ccst), "insert coding constraint box for video tracks", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_ADVANCED},
 	{ OFFS(maxchunk), "set max chunk size in bytes for runs (only used in non-fragmented mode). 0 means no constraints", GF_PROP_UINT, "0", NULL, GF_FS_ARG_HINT_ADVANCED},
--- a/src/filters/mux_isom.c
+++ b/src/filters/mux_isom.c
@@ -293,7 +293,7 @@ typedef struct
 	Bool ssix;
 	Bool ccst;
 	s32 mediats;
-	GF_AudioSampleEntryImportMode ase;
+	u32 ase;
 	char *styp;
 	Bool sseg;
 	Bool noroll, norap;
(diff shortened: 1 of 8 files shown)


touatily:
There is an inconsistency in the code: the argument [ufadts.mpeg2](https://github.com/gpac/gpac/blob/9c261926c5e70647420db9a1ae1f2f085b64b8f1/src/filters/rewrite_adts.c#L397) is an enum but it's declared as a [Bool](https://github.com/gpac/gpac/blob/9c261926c5e70647420db9a1ae1f2f085b64b8f1/src/filters/rewrite_adts.c#L38) and is also treated as a boolean throughout the code.

aureliendavid:
Hi, 

I understand the discussion in #2984 but I feel like changing an explicit type to a generic u32 looses a lot of readability. 

How about a compromise with something like:

```c
typedef enum // NVDecFrameMode
{
	NVDEC_COPY = 0,
	NVDEC_SINGLE,
	NVDEC_GL
} NVDecFrameMode_t ;
typedef u32 NVDecFrameMode; // NVDecFrameMode_t

NVDecFrameMode fmode;
```

it makes it more "grepable" at least.

src/filters/mux_isom.c
```diff
@@ -8464,7 +8464,7 @@ static const GF_FilterArgs MP4MuxArgs[] =
 			"- v1: use v1 signaling, ISOBMFF style (will mux raw PCM as ISOBMFF style)\n"
 			"- v1qt: use v1 signaling, QTFF style\n"
 			"- v2qt: use v2 signaling, QTFF style (lpcm entry type)"
-		, GF_PROP_UINT, "v0", "|v0|v0s|v1|v1qt|v2qt", 0},
+		, GF_PROP_UINT, "v0", "v0|v0s|v1|v1qt|v2qt", 0},
 	{ OFFS(ssix), "create `ssix` box when `sidx` box is present, level 1 mapping I-frames byte ranges, level 0xFF mapping the rest", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_ADVANCED},
```
aureliendavid, line 8468: the first pipe (`|`) is actually not an error

the values are meant to map to the `GF_AudioSampleEntryImportMode` enum whose first value (0) is for `GF_IMPORT_AUDIO_SAMPLE_ENTRY_NOT_SET` (https://github.com/gpac/gpac/blob/master/include/gpac/isomedia.h#L2637)

in order for the default value of the prop (`v0`) to map to the right constant (`GF_IMPORT_AUDIO_SAMPLE_ENTRY_v0_BS = 1`) it needs to be second on the list of possible values 

this is the source of at least some of the test fails of the PR

touatily: Is it normal to have 6 (5 + 1) possible values for this arg. However there are only 5 values in `GF_AudioSampleEntryImportMode enum`?

touatily: Is it normal to have 6 (5 + 1) possible values for this arg. However there are only 5 values in `GF_AudioSampleEntryImportMode` enum?

aureliendavid: hm, in general it's not great but less problematic to have extra values, they will just be ignored

however in this case the extra value is defined, but not it's not in the enum

see: https://github.com/gpac/gpac/commit/1e5ad5bb63bb26529bd04262693cd1f94f4b64e7#diff-e0502e3bb4133c2be5d6f040521b18af03378ede1e3dfb4c7c71910ef2a7136cR39 

I don't know why it's here instead of the enum (ping @jeanlf)  but it exists

aureliendavid:
thank you @gpac-buildbot that was very helpful

src/filters/mux_isom.c
jeanlf: the 6th value (v2qt) is only internal to mp4mux and is defined here https://github.com/gpac/gpac/blob/master/src/filters/mux_isom.c#L40

So yes, it's normal.

jeanlf:
@touatily the `mpeg2` option issue is fixed in master

jeanlf:
thanks!

PR #3025, opened 2024-11-06
Buildbot hintclass

nlsdvl:
Adds GF_ClassTypeHint enum to support clasifying filters in the documentation. 

https://github.com/gpac/wiki/pull/15#issuecomment-2422457046

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -220,9 +220,7 @@ LIBGPAC_FILTERS_FFMPEG=filters/ff_common.o filters/ff_avf.o filters/ff_dec.o fil
 LIBGPAC_FILTERS_LIBCAPTION=filters/dec_cc.o
 LIBGPAC_FILTERS_MPEGHDEC=filters/dec_mpeghdec.o
 
-ifeq ($(CONFIG_EMSCRIPTEN),yes)
 LIBGPAC_FILTERS+=filters/dec_webcodec.o filters/enc_webcodec.o filters/avin_web.o
-endif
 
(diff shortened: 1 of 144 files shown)


DenizUgur:
Were there any braking changes in this PR?

jeanlf:
there should not for gpac, however wiki is now a bit messy -  we're working on it

src/Makefile
```diff
@@ -220,9 +220,7 @@ LIBGPAC_FILTERS_FFMPEG=filters/ff_common.o filters/ff_avf.o filters/ff_dec.o fil
 LIBGPAC_FILTERS_LIBCAPTION=filters/dec_cc.o
 LIBGPAC_FILTERS_MPEGHDEC=filters/dec_mpeghdec.o
 
-ifeq ($(CONFIG_EMSCRIPTEN),yes)
```
rbouqueau, line 223: Is that expected?

DenizUgur: I guess not, there's no use of WebCodecs outside the Web right?

DenizUgur:
> there should not for gpac, however wiki is now a bit messy -  we're working on it

Something broke in the WASM app when querying filters but I couldn't fully debug it.

src/Makefile
jeanlf: it is si that we generate the doc for all platforms

jeanlf:
hmm strange there shouldn't be any change there, I'll look into it

DenizUgur:
Let me debug my wasm build; it's different from the one in this repo.

PR #3029, opened 2024-11-12
fix issues with `--isomedia-only`

DenizUgur:
My goal is to have the smallest build of gpac that only contains isomedia


jeanlf:
Thanks !

PR #3037, opened 2024-11-15
dash client: increase stl tolerance when being ahead of time from 1 seg to 1.5 seg dur

rbouqueau:
The original issue can be reproduced:

```
gpac -i https://livesim2.dashif.org/livesim2/segtimeline_1/testpic_2s/Manifest.mpd dashin:forward=file -o mabr://234.1.1.1:1234 -logs=route@warning -ifce=127.0.0.1 -no-h2
```

and client that will stop playing by lack of seg timeline update:

```
gpac -i mabr://234.1.1.1:1234:repair=no vout -ifce=127.0.0.1
```


rbouqueau:
@jeanlf Good to go!

PR #3038, opened 2024-11-18
Fix --disable-jack, --disable-pulseaudio, --disable-alsa

ntrrgc:
These options in the build were setting the has_jack, has_pulseaudio or has_alsa variables respectively.

However, their values were being later replaced inside the check_has_lib() function, resulting in those options not affecting the enable status of those libraries in anyway.

This patch fixes it by disabling these features with the disabled_packages list, as done for other features.


aureliendavid:
you are correct, thank you for the contrib!

PR #3045, opened 2024-11-22
Refactor enums in structs

touatily:
- Define `GF_OPT_ENUM` macro
- Use `GF_OPT_ENUM` to declare enums in structs for filters

This PR is a follow-up to the discussion we had in the previous PRs #2984 and #3012


rbouqueau:
I like it. Reviewed and so far LGTM. It's been years I had thought of removing these implicit numbers disseminated in the codebase ;)

aureliendavid (approved):
Seems good to me, nice work. 

(and you were right that vscode f12 still works with the macro 👍)

aureliendavid:
I think it's time to merge this otherwise it will diverge too much from master. 

thanks @touatily for the work 

let's be vigilant about whether these changes cause some unpredicted issues, but tests look ok

PR #3046, opened 2024-11-22
Add udta for GF_FilterPacket

DenizUgur:
One use case is if the memory is managed externally and references must be decremented. Without holding a reference to the owner of the memory, no such operation can be done without a workaround.


jeanlf:
Can't you use a packet property for this ?

DenizUgur:
Do you mean something like `GF_PROP_PCK_UDTA` with `GF_PROP_POINTER` type?

DenizUgur:
made the changes

PR #3053, opened 2024-12-10
API Additions for Filter and Constants

DenizUgur:
- **get the filter owning the pid**: If you only have access to the PID you don't have a way to get the owning filter, the new API adds that ability.
- **search stream type by their short name**: This is for easily understanding the stream type just by using the first part of MIME.


DenizUgur:
The tests pass locally; it must be a fluke.

PR #3057, opened 2024-12-14
Add support for encapsulating standalone iamf into mp4

felicialim:


felicialim:
Tests at https://github.com/gpac/testsuite/pull/18

src/media_tools/av_parsers.c
```diff
@@ -23,6 +23,7 @@
  *
  */
 
+#include <include/gpac/internal/odf_dev.h>
```
aureliendavid, line 26: see the lines below, the correct path to include here is 

```c
#include <gpac/internal/odf_dev.h>
```

src/media_tools/av_parsers.c
```diff
@@ -4385,6 +4501,461 @@ GF_Err gf_media_prores_parse_bs(GF_BitStream *bs, GF_ProResFrameInfo *prores_fra
...
+	// LPCM.
+	case GF_4CC('i', 'p', 'c', 'm'):
+		gf_bs_read_int_log(bs, 8, "sample_format_flags");
+		state->sample_size = gf_bs_read_int_log(bs, 8, "sample_size");
+		state->sample_rate = gf_bs_read_int_log(bs, 32, "sample_rate");
+		break;
+	// AAC-LC.
+	case GF_4CC('m', 'p', '4', 'a'):
+		// Read `DecoderConfig`.
+		u32 unused_size;
```
aureliendavid, line 4670: in C99 (which is the standard gpac uses), although you no longer have to declare all variables at the top of the function, you can't have a variable declaration immediately after a goto/case label, because declarations are not considered as statements, and you can only label statements

there's an interesting discussion about it here: https://stackoverflow.com/a/19830820

fix is either to scope the case in a `{}` block, or add a `;` after the label as an empty statement

src/media_tools/av_parsers.c
felicialim: Done.

src/media_tools/av_parsers.c
felicialim: Done, thanks for the link.

src/media_tools/av_parsers.c
```diff
@@ -4385,6 +4501,463 @@ GF_Err gf_media_prores_parse_bs(GF_BitStream *bs, GF_ProResFrameInfo *prores_fra
...
+	}
+	if (num_samples_to_trim_at_start) {
+		*num_samples_to_trim_at_start = read_num_samples_to_trim_at_start;
+	}
+	if (num_samples_to_trim_at_end) {
+		*num_samples_to_trim_at_end = read_num_samples_to_trim_at_end;
+	}
+
+	if (obu_extension_flag) {
+		extension_header_size = (u32)gf_av1_leb128_read(bs, NULL);
```
aureliendavid, line 4604: the reported test timeout https://tests.gpac.io/test/softstretch-rgb/iamf_to_mp4 come from this

here we read a size from the input data, then read data from the input based on this size 

this is a very common scenario for errors, overflows, slow downs, etc. 

when reading sizes from user data we should make sure that the value makes sense before doing anything with it

especially here since this function is called by gf_media_probe_iamf() it can encounter data that is not at all a valid size at this point

checking gf_bs_available() for example would probably be enough here to return an error

there's also a truncation from u64 to u32 here which can lead to wrong values

I haven't looked at the whole PR but in general the pattern of reading a size from user data and then doing malloc()s or reads or writes or pretty much anything based on that size should be taken with a lot of care regarding the validity of the supplied value

src/isomedia/avc_ext.c
```diff
@@ -1419,6 +1419,26 @@ static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg)
 	return out;
 }
 
+static GF_IAConfig* IAMF_DuplicateConfig(GF_IAConfig const * const cfg)
+{
+	u32 i = 0;
+	GF_IAConfig *out = gf_malloc(sizeof(GF_IAConfig));
```
rbouqueau, line 1425: `if (!out) return NULL;`

src/isomedia/avc_ext.c
```diff
@@ -1419,6 +1419,26 @@ static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg)
...
+
+	out->configurationVersion = cfg->configurationVersion;
+	out->configOBUs_size = cfg->configOBUs_size;
+
+	out->configOBUs = gf_list_new();
```
rbouqueau, line 1430: same check

src/isomedia/avc_ext.c
```diff
@@ -1419,6 +1419,26 @@ static GF_AV1Config* AV1_DuplicateConfig(GF_AV1Config const * const cfg)
...
+	for (i = 0; i<gf_list_count(cfg->configOBUs); ++i) {
+		GF_IamfObu *dst = gf_malloc(sizeof(GF_IamfObu)), *src = gf_list_get(cfg->configOBUs, i);
```
rbouqueau, line 1432: same check on dst

src/odf/descriptors.c
```diff
@@ -1935,3 +1932,157 @@ GF_Err gf_odf_opus_cfg_write(GF_OpusConfig *cfg, u8 **data, u32 *size)
...
+
+GF_EXPORT
+GF_IAConfig *gf_odf_ia_cfg_new()
+{
+        GF_IAConfig *cfg = NULL;
+        GF_SAFEALLOC(cfg, GF_IAConfig);
+        if (!cfg) return NULL;
+        cfg->configurationVersion = 1;
+        cfg->configOBUs_size = 0;
+        cfg->configOBUs = gf_list_new();
```
rbouqueau, line 1944: needs to be checked too

src/media_tools/av_parsers.c
```diff
@@ -4385,6 +4501,463 @@ GF_Err gf_media_prores_parse_bs(GF_BitStream *bs, GF_ProResFrameInfo *prores_fra
...
+	GF_IamfObu *a = NULL;
+
+	GF_SAFEALLOC(a, GF_IamfObu);
+	if (!a) {
+		GF_LOG(GF_LOG_ERROR, GF_LOG_CODING, ("[IAMF] Failed to allocate OBU\n"));
+		return;
+	}
+
+	if (!state->bs) {
+		state->bs = gf_bs_new(NULL, 0, GF_BITSTREAM_WRITE);
```
rbouqueau, line 4815: check alloc return

src/filters/reframe_av1.c
```diff
@@ -162,7 +166,10 @@ GF_Err av1dmx_check_format(GF_Filter *filter, GF_AV1DmxCtx *ctx, GF_BitStream *b
 	if (!ctx->state.config)
 		ctx->state.config = gf_odf_av1_cfg_new();
 
-	ctx->is_av1 = ctx->is_vp9 = GF_FALSE;
+	if (!ctx->iamfstate.config)
+		ctx->iamfstate.config = gf_odf_ia_cfg_new();
```
rbouqueau, line 170: check return value on alloc

src/filters/reframe_av1.c
```diff
@@ -379,9 +404,17 @@ static void av1dmx_check_dur(GF_Filter *filter, GF_AV1DmxCtx *ctx)
...
 	}
 	file_size = gf_bs_available(bs);
-	gf_av1_init_state(av1state);
-	av1state->skip_frames = GF_TRUE;
-	av1state->config = gf_odf_av1_cfg_new();
+
+	switch (ctx->bsmode) {
+	case IAMF:
+		gf_iamf_init_state(iamfstate);
+		iamfstate->config = gf_odf_ia_cfg_new();
```
rbouqueau, line 411: idem

src/filters/reframe_av1.c
```diff
@@ -379,9 +404,17 @@ static void av1dmx_check_dur(GF_Filter *filter, GF_AV1DmxCtx *ctx)
...
+		break;
+	default:
+		gf_av1_init_state(av1state);
+		av1state->skip_frames = GF_TRUE;
+		av1state->config = gf_odf_av1_cfg_new();
```
rbouqueau, line 416: idem

rbouqueau:
Thanks!

PR #3058, opened 2024-12-16
Fix segmentation fault when removing an LCT channel

jaunis-ateme:
The iteration on the list was invalid and led to iterate past the last element.


jeanlf:
Thanks !

PR #3066, opened 2024-12-31
Ability to Filter SEIs

DenizUgur:
This PR adds a new option called `seis` to `bsrw`. The new option would allow filtering SEIs rather than removing them all or not (see `rmsei`)

## TODO
- [x] Filter Suffix SEIs for HEVC/VVC
- [x] HEVC rewrite is broken

Testsuite PR: https://github.com/gpac/testsuite/pull/24 https://github.com/gpac/testsuite/pull/25

### DIFF
--- a/src/filters/bsrw.c
+++ b/src/filters/bsrw.c
@@ -705,7 +705,7 @@ static GF_FilterArgs BSRWArgs[] =
 	{ OFFS(pidc), "profile IDC for HEVC and VVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
 	{ OFFS(pspace), "profile space for HEVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
 	{ OFFS(gpcflags), "general compatibility flags for HEVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
-	{ OFFS(seis), "list of SEI message types (4,137,144,...). If used with `rmsei`, this acts as a blacklist; otherwise, it acts as a whitelist", GF_PROP_UINT_LIST, NULL, NULL, GF_ARG_HINT_ADVANCED|GF_FS_ARG_UPDATE},
+	{ OFFS(seis), "list of SEI message types (4,137,144,...). When used with `rmsei`, this serves as a blacklist. If left empty, all SEIs will be removed. Otherwise, it serves as a whitelist", GF_PROP_UINT_LIST, NULL, NULL, GF_ARG_HINT_ADVANCED|GF_FS_ARG_UPDATE},
 	{ OFFS(rmsei), "remove SEI messages from bitstream for AVC|H264, HEVC and VVC", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_UPDATE},
 	{ OFFS(vidfmt), "video format for AVC|H264, HEVC and VVC", GF_PROP_SINT, "-1", "component|pal|ntsc|secam|mac|undef", GF_FS_ARG_UPDATE},
 	{0}
--- a/src/filters/bsrw.c
+++ b/src/filters/bsrw.c
@@ -57,6 +57,7 @@ struct _bsrw_ctx
 	s32 m4vpl, prof, lev, pcomp, pidc, pspace, gpcflags;
 	s32 cprim, ctfc, cmx, vidfmt;
 	Bool rmsei, fullrange, novsi, novuitiming;
+	GF_PropUIntList seis;
 
 	GF_List *pids;
 	Bool reconfigure;
@@ -112,7 +113,7 @@ static GF_Err m4v_rewrite_pid_config(GF_BSRWCtx *ctx, BSRWPid *pctx)
 
 static GF_Err nalu_rewrite_packet(GF_BSRWCtx *ctx, BSRWPid *pctx, GF_FilterPacket *pck, u32 codec_type)
 {
-	Bool is_sei;
+	Bool is_sei = GF_FALSE;
 	u32 size, pck_size, final_size;
 	GF_FilterPacket *dst;
 	u8 *output;
@@ -120,10 +121,8 @@ static GF_Err nalu_rewrite_packet(GF_BSRWCtx *ctx, BSRWPid *pctx, GF_FilterPacke
 	if (!data)
 		return gf_filter_pck_forward(pck, pctx->opid);
 
-
-	final_size = 0;
 	size=0;
-	while (size<pck_size) {
+	while (size<pck_size && !is_sei) {
 		u8 nal_type=0;
 		u32 nal_hdr = pctx->nalu_size_length;
 		u32 nal_size = 0;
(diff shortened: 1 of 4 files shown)


DenizUgur:
```
type 132, size 13, nal_size 18, start 4
type 132, size 13, nal_size 17, start 4
[HEVC] SEI user message type 132 size error (13 but 13 remain), skipping SEI message
```

rbouqueau:
I think it would be more useful with `:nosei=x,y,z` than with `:sei=x,y,z`. Note that we could have both.

DenizUgur:
The issue I see with that is, how do we distinguish if `nosei` was used or not. With uint list you always get `nosei.nb_items==0` if you don't use the option or want to filter out all SEIs.

If you have an idea on that, I'll be happy to change the behavior. We should also not break the current `nosei` behavior.

rbouqueau:
@jeanlf Opinion on how to keep things backward compatible?

jeanlf:
How about using `nosei:seis=1,2` , in which case `nosei` indicates that SEI with types in `seis` list are to be removed, empty list meaning remove everything ?

DenizUgur:
That makes sense. In that case, if `nosei` is not used, then `seis` could be treated as whitelist (current behavior). What do you think?

DenizUgur:
I've updated the branch as you suggested @jeanlf. In this state, `seis` can be used with or without `nosei`.

jeanlf:
Getting there, but for now we only have removal. I think the *_reformat_sei functions should take a boolean is_remove (matching rfnalu.nosei) to enable either dropping the SEI (current behavior) or keeping it. What do you think @DenizUgur ?

DenizUgur:
Do you mean adding a whitelist feature?

DenizUgur:
@jeanlf I've made the change, but I wasn't sure if this is what you meant. The tests will fail, so let me know if this is what you wanted so I can update the tests accordingly.

cconcolato:
> How about using `nosei:seis=1,2` , in which case `nosei` indicates that SEI with types in `seis` list are to be removed, empty list meaning remove everything ?

I find the new syntax a bit too complicated. Having to use `nosei` on top of the other `seis` parameter. Wouldn't it be better to have:
- `nosei` no sei at all
- `sei_incl=1,2` keeps only seis 1 and 2
- `sei_excl=1,2` excludes only seis 1 and 2

I'd be happy with any other names incl or keep, and excl or rem, and any other syntax (camelcase, hyphen or else).

DenizUgur:
That seems cleaner to me as well. @jeanlf @rbouqueau any objections?

jeanlf:
> Here's another idea, probably complicated as well: seis:-4,137 to exclude 4 but include 137.

Not sure how that would work, as you would anyway include other SEIs right ? We want either "remove everything except" or "include only", hence the `:nosei` toggle. What we could do is say that:
- if the first int in the list is negative, all SEIs except the listed ones are removed
- otherwise, all SEIs listed are included

DenizUgur:
Yeah, I haven't given much thought to my suggestion there :)

I'm in favor of Cyril's suggestion if you feel that's okay as well @jeanlf

rbouqueau:
Not sure to understand the use-cases:
1. We currently have a boolean (`:nosei`) removing all SEIs. We want to add the possibility to filter out only certain SEIs. Can we recycle `:nosei`for that?
2. What's the use-case to keep a specific SEI ? I feel modules needing to filter certain packets (e.g. `ccdec`) would do it themselves anyway. I don't imagine people putting a whitelist of SEIs, is that correct?

rbouqueau:
Got a quick chat with @jeanlf . The `:nosei` option comes from MP4Box. Importer stuffs were moved in rfnalu when coding Filters. We should not touch it.

We have another `rmsei` option in [bsrw](https://wiki.gpac.io/Filters/bsrw/) that - to the best of our knowledge - is not used and could be modified.

If we wanted to stay backward compatible we could:
- keep `rmsei` to toggle removal
- add another option to list the SEIs message to either keep or remove

Opinion? CC @cconcolato

cconcolato:
> I don't imagine people putting a whitelist of SEIs

It's exactly the use case I have in mind. You know exactly what SEI have been tested with your players and you only want to keep these ones.

cconcolato:
I'll let you decide what's best given the current behavior. 
My suggestion remains:
>`nosei` no sei at all
>`sei_incl=1,2` keeps only seis 1 and 2
>`sei_excl=1,2` excludes only seis 1 and 2
In this case `nosei` shall not be mixed with `sei_incl` or `sei_excl`.

but if you want to have 
> `nosei` (`seis` not used) remove all SEI
> `nosei:seis=1,2` remove only 1 and 2
> `seis=1,2` (`nosei` not used) keep only 1 and 2
that could work too. The `nosei` is to be interpreted as a modifier of `seis`. 

No strong preference from me.

DenizUgur:
>`nosei` no sei at all
>`sei_incl=1,2` keeps only seis 1 and 2
>`sei_excl=1,2` excludes only seis 1 and 2
> In this case `nosei` shall not be mixed with `sei_incl` or `sei_excl`.

My preference would be this. Seems like the cleanest solution that covers all use cases. If there's no objection I'll revise the PR, update the tests and merge it.

rbouqueau:
Got a chat with @cconcolato . It seems like it could be coded in `bsrw` but mapped directly from a MP4Box option.

So:
- `:nosei` in MP4Box maps to `:nosei` in `reframer` (and it is equivalent to `:rmsei` in `bsrw`)
- `:rmsei` in `bsrw` can be modified as per discussion with @jeanlf earlier today
- the suggestion from @cconcolato and approved by @DenizUgur in the two previous messages

Let's go!

jeanlf:
As said above, this feature should not be in rfnalu but in bsrw, so that we can filter SEIs from e.g. isobmff without having to unframe/reframe, which is a costly operation.

I personally don’t like having 3 mutually exclusive options, so my preference would be to have an `seis` option indicating the list of target SEIs, and `rmsei` indicating if the list is a whitelist or a blacklist. An empty blacklisted list would then mean remove all.

DenizUgur:
Done, testsuite has been updated as well

rbouqueau (approved):
LGTM

src/filters/bsrw.c
```diff
@@ -664,6 +705,7 @@ static GF_FilterArgs BSRWArgs[] =
 	{ OFFS(pidc), "profile IDC for HEVC and VVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
 	{ OFFS(pspace), "profile space for HEVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
 	{ OFFS(gpcflags), "general compatibility flags for HEVC", GF_PROP_SINT, "-1", NULL, GF_FS_ARG_UPDATE},
+	{ OFFS(seis), "list of SEI message types (4,137,144,...). If used with `rmsei`, this acts as a blacklist; otherwise, it acts as a whitelist", GF_PROP_UINT_LIST, NULL, NULL, GF_ARG_HINT_ADVANCED|GF_FS_ARG_UPDATE},
 	{ OFFS(rmsei), "remove SEI messages from bitstream for AVC|H264, HEVC and VVC", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_UPDATE},
```
jeanlf, line 709: could you add that `rmsei` will remove all if seis is empty ?

DenizUgur: done

rbouqueau:
Ok to merge?

PR #3067, opened 2025-01-01
isomedia: implement cslg version 1

bradh:
This adds version 1 support for the CompositionToDecodeBox (`cslg`) as described in ISO/IEC 14496-12:2022 Section 8.6.1.4.

This is used by the GStreamer rust muxer, and is not properly parsed. Instead, there is a command line warning about 20 extra bytes.


rbouqueau:
Thanks! LGTM.

jeanlf:
Thanks !

PR #3074, opened 2025-01-07
Fix compilation warning by adding braces to "if" statement

touatily:
This PR addresses a compilation warning by adding braces { } to "if" statements to properly delimit their blocks.


aureliendavid:
thanks!

PR #3076, opened 2025-01-07
Add `check_stack_size`

DenizUgur:


rbouqueau:
LGTM. Do we want to activate it in the buildbot by default? @jeanlf @aureliendavid

aureliendavid:
as it stands, it leads to the following warnings: 

```
filters/dasher.c:3791:13: warning: stack usage might be 25360 bytes [-Wstack-usage=]
filters/dasher.c:8161:13: warning: stack usage might be 16752 bytes [-Wstack-usage=]
quickjs/quickjs.c:5120:16: warning: stack usage might be unbounded [-Wstack-usage=]
quickjs/quickjs.c:16300:16: warning: stack usage might be unbounded [-Wstack-usage=]
quickjs/quickjs.c:16251:16: warning: stack usage might be unbounded [-Wstack-usage=]
quickjs/quickjs.c:16107:16: warning: stack usage might be unbounded [-Wstack-usage=]
```

the quickjs ones comes from the use of alloca() with a variable size so it will always be triggered with -Wstack-usage

the dasher ones are from dasher_mark_segment_start() and dasher_setup_sources() which allocate a bunch of file paths (GF_MAX_PATH) on the stack

although it's not good practice in general to have a large stack, the danger is pretty vague and the limit in linux is 8MB (far from the 16K default here) - it's mostly a problem when doing recursion or on systems with low max stack size

so I don't know if enabling it by default is of much help (we'd have to add a special case of warning suppression for quickjs), maybe adding it to static/dynamic analysis tests would be better

although I agree that in general we should write smaller functions with smaller stack

rbouqueau:
Let's put 32kb then. Unfortunately, as described in our mailing list discussion, this issue comes from random crashes on arm64 targets on macosx with stack sizes > 64kb (libgpac being used as an API in the original case).

aureliendavid:
oh right sorry I was looking for the rationale behind this change and I had forgotten about that thread

having read it again now it makes more sense

I'd be ok with 32k as I don't feel like re-writing these 2 dasher functions right now (but we can lower it to 16k if someone does)

I'll add a warning exception in the buildbot for the quickjs stuff

aureliendavid:
I've merged it and increased it to 32k for now, we can tweak this later, thanks

aureliendavid:
I had to change Wstack-usage to Wframe-larger-than (which is more or less the same thing I think?) because Wstack-usage doesn't exist on llvm 9.1.0 (used in our macos builder)

PR #3077, opened 2025-01-08
Dump `tx3g` subtitles in inspect

DenizUgur:
Add inspect dump support for `tx3g` and consequently **CEA-708**.

**Example usage:**

```bash
gpac -i cc2.mp4 ccdec inspect:deep:analyze=on
```

**Example output:**

```xml
<PIDConfigure PID="1" name="cc2.mp4" ... StreamType="Text" CodecID="Subtitle/text 3GPP/Apple Stream" ... Codec="tx3g">
 <TextConfig displayFlags="0" horizontalJustification="1" verticalJustification="-1" backgroundColor="0">
  <BoxRecord top="0" left="0" bottom="0" right="0"/>
  <StyleRecord startCharOffset="0" endCharOffset="0" fontID="1" styleFlags="0" fontSize="18" textColor="-1"/>
  <FontRecord fontID="1" name="Serif"/>
 </TextConfig>
</PIDConfigure>
<Packet number="1" PID="1" ...>
<![CDATA[- HEY, CUTIE.
- DARLING.]]>
</Packet>
```


cconcolato:
@DenizUgur When I use the command line above I get only the caption data. How would you use it to have an inspect result that contains both the video inspection result (including timecodes, including SCTE) and the 608 captions, when the input is a TS file?

DenizUgur:
You could try the following to see them together:

```
gpac -i cc2.mp4 ccdec @ @1 inspect:deep:analyze=on
```

CTS of both caption and video frame will match but you may not see them together.

jeanlf (changes requested):

DenizUgur:
@jeanlf do you think anything can be done to interleave two PIDs by their CTS? My previous example doesn't really respect that I think.

jeanlf:
@DenizUgur there's nothing for that in the code and it is dangerous if DTS and CTS differ. We could do this in inspect but this would break the XML which is not designed for interleave.

Is the use case only inspect ?

DenizUgur:
It's not high priority, I was just curious. We need `ccdec` to decode CEA608 but in the use case Cyril mentioned, it would've been useful to see the other messages and properties. The command line I've sent achieves that but the captions and other relevant information are not close to each other.

rbouqueau:
Ok to merge?

rbouqueau:
Thanks!

PR #3078, opened 2025-01-08
uncv: remove duplicate declarations

bradh:


aureliendavid:
thank you, merged

(the test failures above come from the need to merge the upstream head so all good)

PR #3085, opened 2025-01-16
Typos and Refactored code to replace manual character assignments with a call to `strncpy` 

touatily:
Replaced:
  ```c
  data[pos+4] = 'f';
  data[pos+5] = 'r';
  data[pos+6] = 'e';
  data[pos+7] = 'e';
```
with:
```c
strncpy(data + pos + 4, "free", 4);
```


aureliendavid:
on some versions when strncpy doesn't copy the terminal 0 it generates a warning (see https://buildbot.gpac.io//#/builders/12/builds/2004/steps/14/logs/warnings__11_ ) so I changed it to memcpy which does the same thing in this case

PR #3094, opened 2025-01-24
Fix purging LL DASH/HLS with SF mode, as fragment number start from zero

truong-hua:
This fix the bug of purging old segment in dynamic live mode with low latency and separated file enabled


jeanlf:
this is being discussed in #3108

PR #3095, opened 2025-01-25
Make Filter Priority Signed

DenizUgur:
Probably a harmless change, this change could help override an element with a custom one without explicit linking. Since the priority is a `u8` now, it's not possible to go below 0 without `s16`


jeanlf:
Thanks !

PR #3100, opened 2025-02-03
Send encode hints upstream

DenizUgur:
This is required for external applications that needs to encode the incoming stream based on what mp4mx needs. I'm not sure if we need the `last_cdur` or should cover the case where `cdur` can change mid-stream. Is that a sensible case? Should we just send this event once on `mp4_mux_initialize_movie` or `configure_pid`?

### DIFF
--- a/src/filters/mux_isom.c
+++ b/src/filters/mux_isom.c
@@ -198,6 +198,8 @@ typedef struct
 	GF_FilterPacket *dgl_copy;
 	u32 all_stsd_crc;
 
+	GF_Fraction last_cdur;
+
 	Bool has_deps;
 } TrackWriter;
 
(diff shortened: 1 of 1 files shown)


DenizUgur:
Only if `cdur` has changed

rbouqueau:
Who uses `GF_PROP_PID_DASH_FDUR`?

And also there are use-cases like using `dasher:cues`, closing the segment would close the fragment. Is it acceptable?

CC @jeanlf

jeanlf:
I'm having a hard time understanding the use-case here:
- In most cases, fragments are not required to align to GOP boundaries, this will break many deployments
- Sending the event at each change won't guarantee anything. Sending it once at the PID configure would make more sense
- I don't understand why the encoder is not directly configured to produce the desired GOP length, since in your patch the event is only used to signal `cdur` which is a filter option hence is known by the application

DenizUgur:
@jeanlf is there a way to get parsed arguments from a filter session? Like `cdur` that was mapped to `mp4mx`? If that's possible we can close this PR.

rbouqueau:
Maybe check `gpac_check_session_args()` in gpac_help.c

PR #3103, opened 2025-02-12
New workflow to update ABI version automatically

DenizUgur:
The new workflow is designed to help us track the ABI version across commits. The procedure is as follows:  

1. Whenever a push occurs, build both the current commit and the latest `abi-*` tag.  
2. Check if the ABI has changed. If so, determine whether it is a minor or major change.  
3. Update or create the tags to point to the current commit: `abi-<major>.<minor>` and `abi-<major>`.  
4. Open a PR that updates the ABI version inside the code and includes the output of `abipkgdiff` in the PR body.  

**Notes:**  
- In case of staggered pushes, the workflow will be queued and processed atomically.  
- Whenever a new ABI change is pushed, the PR will be updated.  

> [!WARNING]  
> Since we are updating the `abi-*` tags, you must run `git pull --tags -f` to update your local tags.  

> [!IMPORTANT]  
> Before this PR is merged, we need to tag the latest commit with `abi-12` and `abi-12.16`.  

Example screenshots from `git log` and the opened PR:  
log  
pr

### DIFF
--- a/.github/workflows/abi-version.yml
+++ b/.github/workflows/abi-version.yml
@@ -65,7 +65,7 @@ jobs:
         working-directory: gpac_public
         shell: bash
         run: |
-          make distclean
+          make distclean && ./configure --enable-debug
           echo "debian/tmp/usr/include" >> debian/gpac.install
           make deb
           mv -vf gpac*.deb /binaries/
--- a/.github/workflows/abi-version.yml
+++ b/.github/workflows/abi-version.yml
@@ -1,0 +1,236 @@
+name: abi version updater
+on:
+  push:
+    branches:
+      - master
+
+concurrency:
+  group: update-abi
+  cancel-in-progress: false
+
+jobs:
+  prepare-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      latest_abi: ${{ steps.get-tags.outputs.latest_abi }}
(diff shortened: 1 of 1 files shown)


aureliendavid:
Hi, 

Sorry for the delay before getting to it. 

It looks good. I added the tag filter to our scripts everywhere I could find it so that we don't mistake an abi tag for a release tag. 

I tagged the current HEAD to abi-12 and abi-12.6. 

So I think we can merge this and see how it goes. 

(One thing I'm wondering is if we open a PR for each version bump, we need to be quick to merge them so that it doesn't flood the PR requests.) 

Thank you for working on this.

DenizUgur:
The PR will be overwritten so it shouldn't flood it. We should be still quick on it though.

aureliendavid:
Hi again, 

I have a question about this GA: 

it looks like if there are no abi changes in a new push the workflow will update the current abi tags to the latest commit

is that intended? shouldn't the abi tags point to the first commit where a change was detected?

thanks

DenizUgur:
That is intended yes, within the same ABI version the headers and function signatures for libgpac would be same.

The intention behind is that if people would like to pin a specific version of an ABI, they could still receive minor updates like fixes or security updates.

PR #3108, opened 2025-02-12
Delete old fragments in SSR mode

### LINKED ISSUE #3107: dasher does not delete old frags
dasher doesn't seem to delete the old fragments after timeshift buffer depth has passed. I expected the commit https://github.com/gpac/gpac/commit/8936d7c5dfbc2d346e9d315ea37972cef8f53178 to fix it but I think it had some other side-effect that I haven't anticipated

DenizUgur:
closes #3107

### DIFF
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -4733,7 +4733,7 @@ static void dasher_purge_segments(GF_DasherCtx *ctx, u64 *period_dur)
 					u32 k;
 					for (k=0; k<sctx->nb_frags; k++) {
 						char szTmp[100];
-						sprintf(szTmp, ".%u", k + (ds->set->ssr_mode ? 0 : 1));
+						sprintf(szTmp, ".%u", k + (ds->set->ssr_mode || !gf_sys_is_test_mode() ? 0 : 1));
 						char *frag_url = gf_strdup(sctx->filepath);
 						gf_dynstrcat(&frag_url, szTmp, NULL);
 						evt.file_del.url = frag_url;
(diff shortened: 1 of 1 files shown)


rbouqueau:
@jeanlf Can you check that two, it's a two-line change. Thank you!

src/filters/dasher.c
```diff
@@ -4729,11 +4729,11 @@ static void dasher_purge_segments(GF_DasherCtx *ctx, u64 *period_dur)
 				evt.file_del.url = sctx->filepath;
 				gf_filter_pid_send_event(ds->opid, &evt);
 				//purge LLHLS frags
-				if (sctx->frags && (ctx->llhls>1)) {
+				if (sctx->frags && (ctx->llhls>1 || ds->set->ssr_mode)) {
 					u32 k;
 					for (k=0; k<sctx->nb_frags; k++) {
 						char szTmp[100];
-						sprintf(szTmp, ".%u", k+1);
+						sprintf(szTmp, ".%u", k + (ds->set->ssr_mode ? 0 : 1));
```
jeanlf, line 4736: Currently HLS in test mode uses 1-based index and 0-based index otherwise, see https://github.com/gpac/gpac/blob/master/src/media_tools/mpd.c#L4016.
So until we fix this this, I believe the proper patch is

```
sprintf(szTmp, ".%u", k + (ds->set->ssr_mode || !gf_sys_is_test_mode() ? 0 : 1));
```

jeanlf: also the same test needs to be done here https://github.com/gpac/gpac/blob/master/src/filters/dasher.c#L10602

DenizUgur:
@jeanlf here you go

PR #3118, opened 2025-02-24
`avgen` event stream and new `txtgen` filter

DenizUgur:
- **feat: add text stream to avgen**

The new `text` option will send a text frame every 2 seconds. Each line will be read from the lorem ipsum text file.

```bash
gpac avgen:text tx3g2vtt -o live.mpd:dynlast:stl
gpac avgen:dur=5:text -o out.vtt
```

- **feat: add event streams to avgen**

The new `evte` option will send dummy events based on how it's configured.

```bash
gpac avgen:evte=N -o live.mpd # events every Nth second
```


DenizUgur:
Right now it just puts an empty sample at the specified interval. I should probably write an `emsg` box for that.

Do you have any preference on how to configure what event is sent?

rbouqueau:
Should the options be `:type=t` and `:t`to stay consistent with what exists?

The empty evte box is called `emeb`. It's quite similar to `emsg`.

I think we only need a `:type=e` and `:e` for events. The `scte35dec` filter has a `segdur`option which we should leverage if we can.

DenizUgur:
> Should the options be `:type=t` and `:t`to stay consistent with what exists?

I wanted to do it like that but that limits users to enable only text that way. Maybe we can have a `aux=t|e|te` option.

> The empty evte box is called `emeb`. It's quite similar to `emsg`.

Isn't that a sample entry?

> I think we only need a `:type=e` and `:e` for events. The `scte35dec` filter has a `segdur`option which we should leverage if we can.

How do you mean? We just create the pid but don't send any packets through it? Would `scte35dec` create new events at segment boundaries?

rbouqueau:
@cconcolato Feedback welcome

rbouqueau:
@cconcolato Ok to merge?

share/scripts/jsf/avgen/init.js
```diff
@@ -406,9 +434,56 @@ filter.process = function()
...
+{
+	let pck = evte_pid.new_packet(30);
+	pck.cts = evte_cts;
+	pck.dur = filter.fps.n;
+	pck.sap = GF_FILTER_SAP_1;
+
+	//create an empty emsg
+	let bs = new BS(pck.data, true);
+	bs.put_u32(30); //size
+	bs.put_u32(1701668194); //emeb
```
jeanlf, line 453: use put_4cc: https://doxygen.gpac.io/interface_bitstream.html#ada6cad0a5db6cd1fb39654faab0fd9dd

jeanlf (changes requested):

share/scripts/jsf/avgen/init.js
DenizUgur: addressed them

jeanlf:
LGTM, are we ready to merge ?

DenizUgur:
I'm waiting for @cconcolato to go over it, he had some other suggestions for the txtgen filter.

share/scripts/jsf/txtgen/init.js
```diff
@@ -0,0 +1,380 @@
...
+
+filter.pids = [];
+
+filter.set_name("txtgen");
+filter.set_class_hint(GF_FS_CLASS_MM_IO);
+filter.set_desc("Text Generator");
+filter.set_version("1.0");
+filter.set_author("GPAC Team");
+filter.set_help(
+  "This filter generates text streams based on the provided [-src]() file. By default, the filter uses a lipsum text file\n" +
```
cconcolato, line 37:
```suggestion
  "This filter generates text streams based on the provided [-src]() file. By default, the filter uses a lorem ipsum text file\n" +
```

cconcolato (approved):

PR #3126, opened 2025-03-10
Fragmented NHML Streams

DenizUgur:


rbouqueau:
LGTM, thank you!

jeanlf:
Isn't this something that should be handled through a playlist ? I'm a bit worried that we now would allow multi-root XML elements in the context of fragmented NHML, but the XML parser would fail reading a concatenated version of the same stream...

DenizUgur:
My idea behind this was to send `NHNTSample` nodes on-demand from a filter based on some condition or interaction with the user. So, similar to how it's done in ISOBMFF (moov/moof/mdat), I would send `NHNTStream` without any child nodes first to initialize `nhmlr`. Then subsequent `NHNTSample` nodes would be added to the `NHNTStream` root node. In the eyes of `nhmlr` the input XML would only have one root node.

The only expectation here is that each packet should contain a single and complete XML node.

jeanlf:
OK then the NHNTStream potentially never closes, right? 

My concern is that if underlying framing does not respect XML full elements, say for example pushed from a pipe, then we're into trouble. I think the proper way to go is using sax parsing as we do in SVG rather than full file loading as currently done. What do you think ?

DenizUgur:
> OK then the NHNTStream potentially never closes, right?

The assumption is that it would already be arrived in self-closed state. As in: `<NHNTStream ... />`. Subsequent `<NHNTSample />`s would be added to that node. So no need to wait for the closing tag.

> My concern is that if underlying framing does not respect XML full elements, say for example pushed from a pipe, then we're into trouble. I think the proper way to go is using sax parsing as we do in SVG rather than full file loading as currently done. What do you think ?

I haven't used the SAX api yet but if it would mean progressive loading of XML then yes, it makes sense.

DenizUgur:
If I understand correctly, if the input caps accept file only then the XML will be complete regardless if it's piped or not. Can we assume this would only be used externally through libgpac?

PR #3130, opened 2025-03-12
Make sure that `impa` boxes are sorted by item id

maryla-uc:
From the HEIF specification (ISO/IEC 14496-12:2022), Section 8.11.14.1:

> Each ItemPropertyAssociationBox shall be ordered by increasing item_ID, and there shall
> be at most one occurrence of a given item_ID, in the set of ItemPropertyAssociationBox
> boxes.


rbouqueau:
LGTM

PR #3142, opened 2025-03-20
dasher: add error log when no packet forwarded

rbouqueau:
We want to make sure execution fails with `-strict-error`. This did not happen with some faulty stream where GPAC could recover safely, but data integrity was broken.


src/filters/dasher.c
```diff
@@ -10362,6 +10362,9 @@ static GF_Err dasher_process(GF_Filter *filter)
 			}
 			ctx->is_eos = GF_TRUE;
 			gf_filter_pid_set_eos(ctx->opid);
+
+			if (ctx->update_report == 0)
+				GF_LOG(GF_LOG_ERROR, GF_LOG_DASH, ("[Dasher] Error: EOS found but no data sent\n"));
```
jeanlf, line 10367: do not rely on update_report, it could be set back to 0 at that time. It would be cleaner to add a boolean at context level indicating that packets were sent during the period, and resetting this boolean at period switch. If we want to make sure all streams in the period have seen packets, we need to do this at the stream level, and do a loop on period streams while closing the period

jeanlf (changes requested):

src/filters/dasher.c
rbouqueau: Thanks. I've added a new commit to address this.

rbouqueau:
@jeanlf Ready to merge!

PR #3149, opened 2025-03-25
`bsrw` timecode manipulation

DenizUgur:
Add time_code (136) SEI to HEVC/AVC. And equivalent for AV1.

- **force reframer when working with timecodes**
- **move `tc` from `ufobu` to `bsrw` and support removing of tc**

### DIFF
--- a/include/gpac/internal/media_dev.h
+++ b/include/gpac/internal/media_dev.h
@@ -328,6 +328,8 @@ typedef struct
 	Bool remove_video_info;
 	//if set timing info is removed
 	Bool remove_vui_timing_info;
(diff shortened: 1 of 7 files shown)


rbouqueau (approved):
Tested (https://github.com/gpac/testsuite/pull/29). LGTM, thank you @DenizUgur , good work!

DenizUgur:
Thanks @rbouqueau I'm planning on adding drop frame support as well. Afterwards it should be okay to merge

DenizUgur:
This is done. If no further change is needed then I'll merge it after the checks.

PR #3169, opened 2025-04-07
refactor: cleanup timecode handling

DenizUgur:
- **cleanup timecode property**
- **more reliable timecode splitting**
- **apply ts offset based on timecode**

## TODO

- [ ] Parse AV1 metadata and load them to packet properties
- [ ] Require `seiload` from `reframer` if working with timecodes

PR #3178, opened 2025-04-12
`seiload` issues

DenizUgur:
refactored `bsrw` to use `seiload` but encountered bad_access errors from`seiload`.

Added HEVC/AV1 tests to the testsuite so that we can ensure everything is working as required. These hashes were generated with an earlier commit before #3169 was merged.

- gpac/testsuite/pull/30


DenizUgur:
@jeanlf I've started to get segmentation faults from `seiload` when I tried to require it from `bsrw`. I wanted to see if this is some quirk with my Mac's compiler or something else. Let me know if you spot any issues with my diff.

PR #3185, opened 2025-04-15
Improve dasher support for static HDR hevc content

### LINKED ISSUE #3073: DASHer does not properly support SDR BT2020 (cud1) cmaf media profile
The dasher filter adds EssentialProperty and SupplementalProperty to HDR adaptation sets.

However HLG HDR (cud1 cmaf media profile) is assumed as soon as transfer characteristics in the vui is found to be 14 (BT.2020 OETF).

But when the stream is SDR BT.2020 (cud1 cmaf media profile), the transfer characteristics is also expected to be 14, and the dash manifest improperly signals HLG content. For this media profiles, EssentialProperty should also be used to signal color_primaries and matrix_coefficients so that device capability can be queried conveniently.

HLG should be assumed only if the SEI for alternative transfer characteristics is detected.

nlsdvl:
- parse 'alternative transfer characteristic' and propagate it to a new _''ColorTransferAlternative"_ / `GF_PROP_PID_COLR_TRANSFER_ALT` filter property
- update _''ColorTransferAlternative"_  when removing the related SEI with the `bsrw` filter
- write _"EssentialProperty"_ and _"SupplementalProperty"_ to mpd for relevant HDR content

tests : https://github.com/gpac/testsuite/pull/31

fixes #3073


aureliendavid:
(oh boy it looks like a slash in the branch name breaks a lot of scripts and buildbot stuff - I'm taking note of it and will try to spot all the places where we should sanitize the branch name for filenames and such. In the meantime dashes are good, everybody likes dashes)

nlsdvl:
@aureliendavid I closed this one and opened a new one without the slash ... https://github.com/gpac/gpac/pull/3219

Note that using slash in branch names is fairly common, eg. when using 'git flow'.

aureliendavid:
> Note that using slash in branch names is fairly common, eg. when using 'git flow'.

yes you're right, that's my bad

the thing is the buildbot was initially set up to only build master and branches named "buildbot-something" so now that we've extended it to building PRs on demand we're discovering new unexpected things

we'll work to make it more robust

PR #3186, opened 2025-04-16
make utc_ref work with timecodes

DenizUgur:
This PR brings support for using timecodes when `xs` is set to a XML DateTime. Timecodes within media stream are converted to UTC using current UTC date. If the given `xs` value is out-of-bounds for the media stream, then CTS of the packets will reflect that difference. As if the video was playing since `xs` but packets just started to arrive.

## Example

Assume that current UTC time is `2025-04-16T12:16:06Z` (time is not important) and our media stream starts at `14:00:00:00` timecode. The video is running at `60fps`. For simplicity sake there is only video in this example but other streams going through `reframer` will be adjusted so that they follow along the video.

We instruct the reframer to split from `2025-04-16T13:00:00Z` and since timecodes start 1 hour after that time the first CTS will be calculated as follows:

```
1h = 3600s
3600s = 3600 * 60fps = 216000ts
```

First CTS will therefore be **216000**.

```bash
# Create a sample video stream with timecodes starting from 14h
gpac avgen:v:dur=1:fps=60 c=avc:bf=0 bsrw:tc=insert:tcsc=TC14:00:00:00 \
# Split the video stream at 2025-04-16T13:00:00Z, regardless of if first frame is SAP or not
reframer:nosap:utc_ref=tc:xs=2025-04-16T13:00:00Z \
# Inspect the video timestamps
inspect:deep
```

**Output**:
```
PID 1 ID 2 video timescale 60 1280x720 fps 60 SAR 1/1 100 kbps codec avc1.640020 AVC|H264 PL High@3.2 YUV 4:2:0 8 bpp
PID 1 PCK 1 cts 216000 dur 1 sap 1 size 45735
PID 1 PCK 2 cts 216001 dur 1 size 4288
PID 1 PCK 3 cts 216002 dur 1 size 2460
...
```

### DIFF
--- a/src/filters/reframer.c
+++ b/src/filters/reframer.c
@@ -54,6 +54,7 @@ GF_OPT_ENUM (GF_UTCReferenceMode,
 	UTCREF_LOCAL=0,
 	UTCREF_ANY,
 	UTCREF_MEDIA,
+	UTCREF_TC,
 );
 
 
@@ -89,7 +90,6 @@ typedef struct
 
 	u64 ts_at_range_start_plus_one;
 	u64 ts_at_range_end;
-	u64 ts_tc_offset;
 
 	GF_List *pck_queue;
 	//0: not computed, 1: computed and valid TS, 2: end of stream on pid
@@ -175,6 +175,7 @@ typedef struct
 	GF_TimeCode *cur_start_tc, *cur_end_tc;
 	Bool cur_start_valid, cur_end_valid;
 	u64 start_frame_idx_plus_one, end_frame_idx_plus_one;
+	GF_Fraction64 ts_tc_offset;
 
 	Bool in_range;
 	Bool load_sei;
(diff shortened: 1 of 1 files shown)


rbouqueau:
Code LGTM. I think I understand the use-case for this.

IIUC setting `:utc_ref=tc` changes `xs` from "extraction within range" to "extraction whatever" (by offsetting the cts), I don't find this particularly obvious both as a behaviour and from the documentation.

In practice I would have expected 2 steps: 1) offset the timecode and 2) make the extraction (but do we really make an extraction here?). Alternately we could rename `:utc_ref=tc` to `:utc_ref=tc_reset` or `:utc_ref=tc_offset` to make more explicit that there is an operation involved.

Opinion?

DenizUgur:
This is just an example when `xs` is before the first UTC reference. The same setup works if the UTC ref is in future.

rbouqueau:
Well what is supposed to happen is UTC ref is in future? The cts becomes negative, correct?

rbouqueau (approved):

rbouqueau:
Thank you

PR #3219, opened 2025-04-28
Improve dasher support for static HDR hevc content

### LINKED ISSUE #3073: DASHer does not properly support SDR BT2020 (cud1) cmaf media profile

nlsdvl:
- parse 'alternative transfer characteristic' and propagate it to a new _''ColorTransferAlternative"_ / `GF_PROP_PID_COLR_TRANSFER_ALT` filter property
- update _''ColorTransferAlternative"_  when removing the related SEI with the `bsrw` filter
- write _"EssentialProperty"_ and _"SupplementalProperty"_ to mpd for relevant HDR content

tests : https://github.com/gpac/testsuite/pull/31

fixes #3073


rbouqueau:
Thank you!

PR #3220, opened 2025-05-03
Parse command-line through libgpac APIs

DenizUgur:
The PR adds two methods that almost copy what gpac binary does to parse the command line. It is a bit simpler than what gpac does but should be enough for most applications using libgpac.

`process_link_directive` method is identical to what we have on `gpac.c` so maybe that could be moved to somewhere more common but no idea where would be a good place for it.


jeanlf:
Interesting - any thoughts on using these new functions in gpac.c to avoid code duplication ?

DenizUgur:
I was doing that initially but there are too many dependencies to the gpac options which makes the function signature crowded. In reality, we wouldn't need the exact same level of complexity in filter graph compared to gpac command line.

Only thing I would've preferred to de duplicate is the `process_link_directives` function. What do you think?

DenizUgur:
From my understanding `GF_FS_DEFAULT_SEPS[5]` equals to `@` and that's always same, isn't?

https://github.com/gpac/gpac/blob/d133fadaaf604ca67a1a9e86c2204d3de912295d/src/filter_core/filter_session.c#L620-L626

I've de-duplicated the `process_link_directive` function as well.

applications/gpac/gpac.h
```diff
@@ -49,6 +49,8 @@ Bool mp4c_handle_prompt(u8 char_val);
 Bool mp4c_task(void );
 void mp4c_help(u32 argmode);
 
+GF_Err process_link_directive(char *link, GF_Filter *filter, GF_List *loaded_filters, char *ext_link);
+
```
jeanlf, line 53: You should:
- rename to gf_fs_process_link_directives
- add the exported symbol in src/export.cpp

jeanlf (changes requested):

jeanlf:
> From my understanding GF_FS_DEFAULT_SEPS[5] equals to @ and that's always same, isn't?

Yes because until this PR, this was only used for gpac parsing of link directives, not used in the filter session. I think we can keep it harcoded to '@' for the time being.

applications/gpac/gpac.h
DenizUgur: This is not going to be consumed by users of libgpac right? I haven't included this in `include/filters.h` either. So should we still rename it and export it from `export.cpp`?

DenizUgur:
> Yes because until this PR, this was only used for gpac parsing of link directives, not used in the filter session. I think we can keep it harcoded to '@' for the time being.

As you wish, it should be easy to fix that here though.

applications/gpac/gpac.h
aureliendavid: if it's defined in the lib (filter_session here) and used in an executable (gpac) it should be exported by the lib so we can link the binaries dynamically 

see the build fails on windows: https://buildbot.gpac.io//#/builders/13/builds/2106/steps/12/logs/stdio

DenizUgur: Fair enough, I'll export it from public headers as well. Do we need to modify any other file to export it correctly?

aureliendavid: the GF_EXPORT on the definition in filter_session.c handles it for all other OSes (the builds look ok)

adding it to export.cpp for windows should be enough (windows/msvc is pretty annoying with this)

PR #3221, opened 2025-05-04
Compatibility with C++14

DenizUgur:
Recently, I tried to compile an example application that inputs and outputs data using GPAC in C++. The only compatibility issue I observed was with named designators.

### DIFF
--- a/include/gpac/filters.h
+++ b/include/gpac/filters.h
@@ -1620,61 +1620,61 @@ u8 gf_props_4cc_get_flags(u32 prop_4cc);
 
 
 /*! Helper macro to set signed int property */
-#define PROP_SINT(_val) (GF_PropertyValue){.type=GF_PROP_SINT, .value.sint = _val}
+#define PROP_SINT(_val) (GF_PropertyValue){GF_PROP_SINT, { .sint = (_val) }}
 /*! Helper macro to set unsigned int property */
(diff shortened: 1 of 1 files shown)


rbouqueau:
Loving it, thanks! Anyone has any remark before I merge?

aureliendavid:
can I ask which compiler you are using? (or maybe if you can share in private the whole example tested?)

I'm a bit confused about things like: 

```c
#define PROP_FRAC_INT(_num, _den) (GF_PropertyValue){GF_PROP_FRACTION, { .frac.num = (_num), .frac.den = (_den) }}
```

is this allowed in c++14?

Also aren't GF_FilterRegister declarations not an issue? (with lots of named designators)

DenizUgur:
I'm compiling the application with C++, not libgpac so only public exports needs to be compatible with C++14.

The compiler seems to be from Ubuntu not fully sure it's GNU though (it doesn't specify in `--version`)

rbouqueau:
What else should we test here? It would be damn convenient to be able to use the GPAC C headers from C++ if some SWIG bindings are generated from the C headers already.

rbouqueau:
Wouldn't it make sense to upload the `memio` example in `share/` with some other bindings? And then test it in our CI/CD?

jeanlf:
Thanks !

PR #3222, opened 2025-05-05
Preserve timestamp for RAW streams

DenizUgur:


src/filters/reframer.c
```diff
@@ -1774,8 +1778,15 @@ GF_Err reframer_process(GF_Filter *filter)
 
 			const GF_PropertyValue *p = gf_filter_pid_get_property(ipid, GF_PROP_PID_CODECID);
 			u32 codec_id = p ? p->value.uint : GF_CODECID_NONE;
-			if (codec_id != GF_CODECID_AVC && codec_id != GF_CODECID_HEVC && codec_id != GF_CODECID_AV1)
-				continue;
+			switch (codec_id) {
+				case GF_CODECID_AVC:
+				case GF_CODECID_HEVC:
+				case GF_CODECID_AV1:
+				case GF_CODECID_RAW:
```
rbouqueau, line 1785: So weird to not see VVC from the removed code, and to see GF_CODECID_RAW in the added one. May be right but looks unbalanced at first sight.

DenizUgur: Honestly this entire check is a bit redundant, if the packet doesn't have timecode then rest won't be run anyway. This check limits streams that could contain timecode property.

PR #3224, opened 2025-05-07
Split stream at segment boundary in reframer

DenizUgur:
When we give raw frames to reframer, the concept of SAPs don't really make sense. However, if downstream expects to receive SAPs at deterministic intervals we need to ensure we start correctly. It's up to the downstream encoder to ensure GOP length is fixed afterwards.


rbouqueau:
LGTM. Do we have some tests?

DenizUgur:
Nope, do we need any?

rbouqueau:
Well I guess at some point this workflow needs to be document and tested to avoid regressions.

jeanlf:
this looks like duplicating of `dasher.segdur` no ? A better way would be to catch `GF_FEVT_ENCODE_HINTS` and get the info from the event - note that this would require not canceling this event in the filters processing it, which I believe is OK.

DenizUgur:
My initial approach was like that but I didn't want to cancel events in the encoder. Only downside this would have is that, if there are two encoders in upstream they would both receive the same encode hints. Although unlikely, this might produce a bit of unexpected behavior.

jeanlf:
The problem with an explicit option is that a user could set it to a different value than the target dash segment duration, giving invalid results. I'm not sure I understand why you don't like using the encode hints event, could you elaborate ?

DenizUgur:
I would prefer encode hints. My concern is that, that encode hints would go to all encoders in the upstream not just the encoder. Only solution I see is that we could apply the encore hints event to an encoder only if `fintra` is not already specified.

DenizUgur:
Currently, this behavior is the default for raw streams with a downstream dasher. It might be sensible to introduce a boolean option to enable or disable it.

jeanlf (changes requested):
We're getting there but I don't think we should create a new event. My suggestions are:
- rename GF_FEVT_ENCODE_HINTS to GF_FEVT_TRANSPORT_HINTS
- add a 32 bit flags field in the associated structure, and define one flag to indicate 'encoder has seen this' (so that encoders towards the source can ignore the event)
- cancel the event in ffenc and dec_scte35, and send a new event of same type with the bit flag set

Seems reasonnable ?

DenizUgur:
All seems reasonable. Modifed as you suggested,

rbouqueau:
I'm still checking and working on this one here: https://github.com/rbouqueau/gpac/tree/buildbot-reframer-sdsap

rbouqueau:
The current test fail is due to some https://github.com/gpac/gpac/pull/3239 / https://github.com/gpac/testsuite/pull/33 commits being integrated while not merged.

rbouqueau:
Some remarks after a batch of checks I did:
- If the UTC xs time doesn't correspond to a video frame, the dasher put an offset on the video (which can be seen with `:stl` by comparing the first entry with the pto). A priori 1) the time should correspond to a video frame (which means all frames timestamps are expressible) and 2) we may want the reframer to be video-driven? 
- Subtitles are well resegmented.
- Event tracks may miss the media segments.

PS: `gpac avgen:dur=10:evte=1 txtgen:type=utc:fdur=5:lock:dur=10 @@1 tx3g2vtt @@1 tx3g2ttml c=avc:fintra=1:bf=0 @ bsrw:tc=utc c=aac reframer:rt=sync:utc_ref=tc:xs=0 @ -o dash/gpac.mpd:stl:segdur=2:rawsub`

PS2: I confirm there is an issue with the delay in the dasher:
- `gpac avgen:dur=1:evte=1 restamp:delay=1 c=aac c=avc -o dash/gpac.mpd:stl` (or `gpac avgen:dur=10:evte=1 c=aac c=avc:bf=0 restamp:delay=1 -o dash/gpac.mpd:stl`)
- `gpac -i dash/gpac.mpd inspect:deep`

rbouqueau:
As a reminder, the command-lines that serves for testing at this point:

```
gpac avgen:ntp:dur=10:evte=1 txtgen:type=utc:fdur=1:lock:dur=10 @@1 tx3g2vtt @@1 tx3g2ttml c=avc:bf=0 @ bsrw:tc=utc c=aac -o 10s.mp4:frag
gpac -i 10s.mp4 reframer:utc_ref=tc:xs=xxxxx -o /tmp/dash/gpac.mpd:segdur=2:stl:timescale=1000
```

As a reminder, as a one-liner:
```
gpac avgen:ntp:lock:dur=10:evte=1:FID=GEN txtgen:type=utc:fdur=1:lock:dur=10 @@1 tx3g2vtt:FID=S1 @@1 tx3g2ttml:FID=S2 c=avc:bf=0:fintra=2 @ bsrw:tc=utc:FID=V1 c=aac:FID=A1 reframer:SID=S1,S2,V1,A1,GEN#event:FID=REF c=aac:FID=AA1 c=avc:bf=0:fintra=2:FID=VV1 -o /tmp/dash/gpac.mpd:segdur=2:stl:timescale=1000:segcts:profile=live:mvex:sdtp_traf=both:tfdt_traf:straf:nofragdef:rawsub:SID=AA1,VV1,REF#event,REF#text -graph
```

Another one:
```
gpac -lu -no-dynf -full-link -graph --dur=10 --stl --timescale=1000 \
avgen:sr=48000:sizes=960x540:ntp:lock:evte=1:adjust:fps=25:dyn=false:banner=ftab1:#ID=\(event\)1:#Representation=\(event\)884342069:#ASID=\(event\)884342069:FID=AVGEN \
txtgen:type=utc:fdur=1:lock:FID=TXTGEN \
ffsws:osize=768x432:SID=AVGEN#video @ c=avc:bf=0:b=750k:fintra=2:fps=25 @ bsrw:tc=utc:FID=avc0:#ID=1:#ASID=884337431:#Representation=884337431 \
c=aac:b=96k:SID=AVGEN#audio:FID=aac1:#ID=1:#ASID=884342785:#Representation=884342785 \
tx3g2ttml:SID=TXTGEN:FID=ttml2:#ID=1:#ASID=884341132:#Representation=884341132 \
tx3g2vtt:SID=TXTGEN:FID=vtt3:#ID=1:#ASID=884341267:#Representation=884341267 \
reframer:rt=sync:SID=avc0,aac1,ttml2,vtt3,AVGEN#event \
@ -o /tmp/dash/gpac.mpd:segdur=2:segcts:profile=live:dmode=dynamic:tsb=30:mvex:sdtp_traf=both:tfdt_traf:straf:nofragdef:rawsub:template=\$RepresentationID\$/s_\$Init=init\$\$Segment=\$\$Number\$
```

rbouqueau:
I've just updated the failing hash, after checking that all was fine.

rbouqueau:
That branch contains improvements that may conflict with future devs. Can we merge?

jeanlf:
thanks

Issue #3229, opened 2025-05-14
TTML output missing region reference


cconcolato:
In this code, a TTML region is defined but never referenced:
https://github.com/gpac/gpac/blob/f8fa51f721fd7530e3a18d2e5a7594e122e65044/src/filters/write_tx3g.c#L226

I suggest following the example in https://www.w3.org/TR/2018/REC-ttml2-20181108/#semantics-default-region and adding `region="Default"` to the `body` element. Alternatively, you could add the `region="Default"` attribute to all `p`.

nigelmegitt:
Friendly reminder: if you apply a default region at the body level and then a descendant specifies a different region, that descendent and its whole sub-tree will be pruned by the TTML processor and never presented. I'm not sure if that's relevant in this case.

PR #3232, opened 2025-05-15
ttml: reference default region in body

rbouqueau:
Trying to solve #3229. Is there anything else we may want to add here?


rbouqueau:
How can I relaunch the tests here now that I've updated the testsuite?

rbouqueau:
Ready to merge. Don't forget to merge https://github.com/gpac/testsuite/pull/32 as well :)

PR #3233, opened 2025-05-16
Fix IAMF-FLAC sample size and sample rate.

jwcullen:
- Fixes unexpected timescales and possible AV sync issues in IAMF-FLAC.
  - Sample rate is 20 bytes in section 8.1 of RFC 9639.
  - The coded sample size is offset by 1.


rbouqueau:
I checked, it's perfect. Thank you!

PR #3237, opened 2025-05-21
Allow to escape list separator by doubling it

rbouqueau:
The [wiki](https://wiki.gpac.io/Filters/filters_general/#escaping-option-separators) states that "When a filter uses an option defined as a string using the same separator character as gpac, you can either modify the set of separators, or escape the separator by duplicating it.".

This didn't work with `httpout:hdrs` when the HTTP header value contained a comma (default separator). This patch discard the separator if it is immediately followed by the same separator.

Should we check for other cases where it could not be handled properly?


jeanlf:
LGTM - could you add some text in https://github.com/gpac/gpac/blob/master/applications/gpac/gpac_help.c#L159 to indicate that comma can be escaped in lists with an example ?

rbouqueau:
@aureliendavid Regression tests couldn't execute properly due to download issues on the content server. Would it be possible to re-launch tests once this is solved? Thanks

rbouqueau:
@aureliendavid Thx. @jeanlf Ready to merge!

jeanlf:
thanks!

PR #3281, opened 2025-07-04
Support Dolby AC-4

MingxinChen:
1. Add AC-4 muxing and demuxing features
2. Fix the channel count bug in AC-3/EC-3


jeanlf:
thanks for this ! Do you have public test content that we could use in our test suite ?

jeanlf:
now in masyter, I found a sample ac4 stream for the test suite - many thanks !

PR #3309, opened 2025-07-22
add new workflows for publishing apt repo and wasm binaries

DenizUgur:


aureliendavid:
Looks good to me. 

We can test them during the slow summer period to see if it all goes well. 

(as we discussed earlier maybe we can add the gpac.html file to the wasm distrib) 

thanks for the work!

PR #3318, opened 2025-07-30
add the ability to use AVFormatContext as source via ffdmx

DenizUgur:


jeanlf:
Isn't this the same as using ffdmxpid filter ?

DenizUgur:
How so? The intend behind this PR is to use libgpac alongside libav and pass the pointer of AVFormatContext as source to gpac. How do you achieve that with `ffdmxpid`?

rbouqueau:
@jeanlf @DenizUgur What's the status on this?

DenizUgur:
If it's not a redundant work, then this PR is done on my side.

jeanlf:
This seems quite dangerous to me: we have no clue of checking that the ABI of the external libavformat matches what is used in ffdmx which will likely lead to crashes ...

DenizUgur:
That's true, do you have any suggestions? This is working quite well for me and I would be happy to make it as safe as possible.

jeanlf:
I don't have a clean solution for that no. Maybe we could do some checks on demuxer->av_class->version ?

jeanlf:
OK that should do it for a first version

PR #3343, opened 2025-09-19
Event Track passthrough

rbouqueau:
Even Tracks can be used for data that can't be parsed e.g. custom event tracks, nhml inputs, or data that GPAC can't handle. Tests will be added to the non-reg testsuite.

### DIFF
--- a/include/tests.h
+++ b/include/tests.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <gpac/tools.h>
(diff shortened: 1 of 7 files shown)


rbouqueau:
I clicked on the wrong button, this is still a draft. I edited the title to make it clear.

rbouqueau:
All green. Merging.

PR #3346, opened 2025-09-24
add av1 switching frames 'av1s' sample group

rbouqueau:


rbouqueau:
@jeanlf Tests are broken because the SAP type encoding is broken. We mix ISOBMFF SAPs with the `GF_FilterSAPType` type. Ok to fix that or do you envision something different for AV1 switching frames?

src/isomedia/box_funcs.c
```diff
@@ -1174,6 +1174,8 @@ static struct box_registry_entry {
 	SGPD_DEFINE( GF_ISOM_BOX_TYPE_SGPD, sgpd, "stbl traf", GF_ISOM_SAMPLE_GROUP_3GAG, "3gpp"),
 	SGPD_DEFINE( GF_ISOM_BOX_TYPE_SGPD, sgpd, "stbl traf", GF_ISOM_SAMPLE_GROUP_AVCB, "3gpp"),
 
+	SGPD_DEFINE( GF_ISOM_BOX_TYPE_SGPD, sgpd, "stbl traf", GF_ISOM_SAMPLE_GROUP_AV1S, "av1-isobmff"),
```
jeanlf: should be "av1" not "av1-isobmff" as for other AV1 boxes

rbouqueau:
@cconcolato I've added support to segment at cues.

@jeanlf I've duplicated export `gf_isom_get_sample_rap_roll_info()` into non-exported `isom_get_sample_switch_frame()` to avoid an API dump. I feel I should have merged them, what do you think?

jeanlf:
Thanks !

PR #3348, opened 2025-09-30
dash ondemand: don't write empty moof

rbouqueau:
reproduced with dual track content


jeanlf:
I feel that's a bit too harsh - if people want to use the lib to generate empty moofs with emsh or ... we shouldn't stop that from happening. It would be better to fix the root cause in mp4mx.

rbouqueau:
I'll provide to you the details. Maybe this could be made an option?

DenizUgur:
Also related to this PR, the `flist` filter with `floop=-1` seems to trigger this as well. I'm seeing empty `moof` boxes whenever the file list loops.

rbouqueau:
Better now? @jeanlf

jeanlf:
Yes, thanks !

PR #3352, opened 2025-10-02
cmaf: fix condition to remove mehd

rbouqueau:


rbouqueau:
@jeanlf If that's ok I'll update the hash of the encryption test that uses `:cfmc` as well.

Murmur:
How do `remove_mehd` flag works? We use MP4Box to encode a static dash vod content from input.mp4 files, so a total duration is known by the mp4box. 
Files(mp4, m4s, mpd) are copied to a Livesim2 asset folder, run a java tool to modify `init.mp4 -> init_livesim.mp4` to remove `moov/mvex/mehd` table, livesim manifests point to a new init_livesim.mp4 file.
* we also deliver the same test content as a static dash so want to reuse everything as much as possible.

How do this fit in our test material pipeline? 

ps: I also remove `moov/udta` table from init.mp4 files (vod and live delivery for hbbtv test material), it used to raise some weird problem in some HbbTVs so I still just remove it. It should not raise side effects but it did for unknown reason. I probably should keep `udta` in a new test data and let manufacturers deal with it.

rbouqueau:
@Murmur That's automatic when specifying a cmaf profile.

jeanlf:
Thanks !

aureliendavid:
can you update test or hash for `encryption-segment`? (see https://tests.gpac.io/testres/test/encryption-segment/master)

thanks!

rbouqueau:
@aureliendavid Sorry I forgot to merge https://github.com/gpac/testsuite/pull/41 when this was merged.

PR #3353, opened 2025-10-04
MP4Box, fix FPS with `AddChapter(nb_frames,chapter name)`

andersstorhaug:
I was using the MP4Box `-chap` option with ZoomPlayer's `AddChapter(nb_frames,chapter name)` (frame) syntax, and I noticed that I had to supply `-fps 1001000/60` to make it work correctly for 59.94 FPS (which, was correctly detected).

Looks like the issue is that the arguments are flipped, which makes some sense, given that $\frac{1}{\frac{1001000}{60}} \times 1000 \times 1000 = 59.941$.


rbouqueau:
PS: the mem leak from the buildbot is not present on master

andersstorhaug:
Rebased to upstream `master`.

jeanlf:
Thanks !

PR #3374, opened 2025-10-21
ffavf: handle input fps

rbouqueau:
fixes #3371

PR #3376, opened 2025-10-22
Merge properties in ffavf if graph is 1:1

### LINKED ISSUE #3372: Properties are not merged in `ffavf`
I assume by design properties are not merged between two ends of `ffavf` but this results in `SenderNTP` being lost. For simple filter graphs (1 in, 1 out) we may merge the properties as it's done in `ffsws` or `ffenc`.

One use case for this is to use `drawtext` filter to draw the current system clock while preserving the sender ntp for glass-to-glass latency measurements.

DenizUgur:
fixes #3372

### DIFF
--- a/src/filters/ff_avf.c
+++ b/src/filters/ff_avf.c
@@ -124,10 +124,14 @@ static GF_Err ffavf_setup_input(GF_FFAVFilterCtx *ctx, GF_FFAVPid *avpid)
 	}
 	//destroy filter (will remove from graph)
 	if (avpid->io_filter_ctx) avfilter_free(avpid->io_filter_ctx);
-	avpid->io_filter_ctx = NULL;
-	ret = avfilter_graph_create_filter(&avpid->io_filter_ctx, avf, pid_name, args, NULL, ctx->filter_graph);
-	if (ret<0) {
-		GF_LOG(GF_LOG_ERROR, GF_LOG_MEDIA, ("[FFAVF] Fail to create filter graph: %s\n", av_err2str(ret) ));
+	avpid->io_filter_ctx = avfilter_graph_alloc_filter(ctx->filter_graph, avf, pid_name);
+	if (!avpid->io_filter_ctx) {
+		GF_LOG(GF_LOG_ERROR, GF_LOG_MEDIA, ("[FFAVF] Fail to create filter graph\n"));
+		return GF_OUT_OF_MEM;
+	}
+	ret = avfilter_init_str(avpid->io_filter_ctx, args);
+	if (ret < 0) {
+		GF_LOG(GF_LOG_ERROR, GF_LOG_MEDIA, ("[FFAVF] Fail to initialize filter graph: %s\n", av_err2str(ret) ));
 		return GF_BAD_PARAM;
 	}
 	return GF_OK;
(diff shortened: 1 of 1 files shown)


rbouqueau:
LGTM but will conflict with #3374. Do we know what diverges in the two changes from the testsuite? (which may also conflicts with https://github.com/gpac/testsuite/pull/42)

aureliendavid:
A way to go would be to: 

- decide which PR takes precedence between this one and https://github.com/gpac/gpac/pull/3374
- fix the tests for the priority 1 PR until the buildbot is happy
- merge the first PR
- rebase the second PR to new master
- fix the tests until buildbot is green
- merge the second PR

DenizUgur:
@rbouqueau I haven't tried it but I wonder if this PR indirectly fixes your PR as well.

**EDIT:** It doesn't fix it directly. But I would assume if we don't override timescale and fps then it might be okay.

rbouqueau:
@DenizUgur Are you ok with me merging my branch in yours here?

DenizUgur:
Yes, go ahead.

rbouqueau:
Something is not right. @DenizUgur Can I ask you to test the tests on the eponym testsuite branch? It works here but not on the testbot.

aureliendavid:
> Something is not right. @DenizUgur Can I ask you to test the tests on the eponym testsuite branch? It works here but not on the testbot.

I'm seeing the same thing as the buildbot for the [ffavf-a2a-cfg](https://tests.gpac.io/test/ffavf-a2a-cfg/buildbot-ffavf-merge-props) test. 

On the branch its result hashes to `79446A0787ECA17DA8269AB3A6206C2153F1D3F1` 

but the ref hash in the testsuite PR at https://github.com/gpac/testsuite/pull/43/files#diff-8752109f9c590b336962033bd1616b76cf5f4fb280a9f2c124b10ad6dacf21b5 was changed to `BFE70F8AFC77BC3711C5CBB7C5435D358DD8A0F7` in the second commit (https://github.com/gpac/testsuite/pull/43/commits/71fa65c54a637d63ef607504e94c1e9bce96dab0) 

Here's the resulting file: [inspect.txt](https://github.com/user-attachments/files/23092223/inspect.txt)

You can check if the test produces a different file for you. Maybe from different versions of ffmpeg?

rbouqueau:
> You can check if the test produces a different file for you. Maybe from different versions of ffmpeg?

Good catch. Now trying to understand where it comes from.

rbouqueau:
I'm on FFmpeg 8.0.1:

```
% gpac -i media/auxiliary_files/enst_audio.aac ffavf:afmt=s16:sr=44100:dump::f=acompressor @ inspect:interleave=false:deep:fmt=%pn%-%dts%-%cts%-%sap%-%size%%lf%:
[out] Option 'sample_fmts' is not a runtime option and so cannot be set after the object has been initialized
[FFAVF] Fail to set abuffersink audio format: Invalid argument
[out] Option 'sample_rates' is not a runtime option and so cannot be set after the object has been initialized
[FFAVF] Fail to set abuffersink audio sample rate: Invalid argument
[FFAVF] Graph dump:
...
```

I'll follow up later on this.

jeanlf:
looks good to me thanks - please update tests before merging

PR #3379, opened 2025-10-28
check port before attaching parent http session

nlsdvl:
Not checking the port before attaching parent http session causes unexpected and confusing behaviors, eg. in a local development configurations where the server name is the same and only the port varies.

cc @rbouqueau


rbouqueau:
Thanks, LGTM. @jeanlf ?

jeanlf:
Thanks !

PR #3390, opened 2025-11-11
Support Query Parameters in DASH

DenizUgur:
This addition is described in ISO/IEC 23009-1 (Annex I). It does not cover all the features but should be sufficient for most cases.


rbouqueau:
LGTM, thank you! CC @jeanlf ok with that?

jeanlf:
Thanks !

PR #3399, opened 2025-11-25
ttml: rewrite timestamp on multi periods

rbouqueau:
`gpac flist:srcs=s.ttml:floop=1:sigperiods -o out.mpd:ttml_agg:segdur=1`

### DIFF
--- a/src/filters/write_generic.c
+++ b/src/filters/write_generic.c
@@ -951,6 +951,7 @@ static u64 ttml_get_timestamp(char *value)
 	return ttml_get_timestamp_ex(value, 25, &ttml_fps_num, &ttml_fps_den, &ttml_sfps);
 }
 
+// modifications in this function should be mirrored in load_text.c-ttml_rewrite_timestamp()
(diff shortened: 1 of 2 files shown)


rbouqueau:
I repushed  the test because I had made a typo while editing the test - my bad.

src/filters/write_generic.c
```diff
@@ -937,6 +941,57 @@ static GF_Err ttml_embed_data(GF_XMLNode *node, u8 *aux_data, u32 aux_data_size,
 	return GF_OK;
 }
 
+
+//from filter txtin
```
jeanlf: we should avoid duplicating these - maybe add generic ttml timestamp rewrite in src/media_tools/mdia_import.c ?

jeanlf: Fine with me - just add a note also in load_text that the any changes in the ttml rewrite code should be reflected in write_generic

rbouqueau:
@jeanlf Ready to go!

jeanlf:
thanks!

PR #3402, opened 2025-11-27
Long run issue commits

rbouqueau:
These commits allowed me to avoid early exits saw in some long runs with complex avgen -> enc -> reframer -> dec -> enc -> reframer , where the reframer gave the impression to provoke an EOS.


aureliendavid:
seems good to me

jeanlf:
Thanks!

PR #3404, opened 2025-11-28
Fix ABI version updater workflow

DenizUgur:
ABI version updater workflow has been dormant since c72bcdfb3b5572ad91f3c80db58e68599303173a because `abipkgdiff` doesn't register SONAME changes as "incompatible ABI" and since that commit has bumped ABI major version, it stopped working.

04ccdf47f35cf439a198aaf1213180347b292354 has changed the way we bump ABI versions to be automatic. And since the ABI difference detection is pretty good (when you use the tool properly...) I believe it's time we make this automatic for good.

This PR fixes some issues from that commit and changes the tool we use (still from `abigail-tools`) to `abidiff`. This one uses the `.so` instead of the debian package.

When we merge this PR, any future ABI changes will be commited directly after that particular change. The body of the commit will contain the output of `abidiff`. The pending ABI 14.0 version will have the following body for instance: [abi-diff.txt](https://github.com/user-attachments/files/23814466/abi-diff.txt)

### DIFF
--- a/.github/workflows/abi-version.yml
+++ b/.github/workflows/abi-version.yml
@@ -14,12 +14,13 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       latest_abi: ${{ steps.get-tags.outputs.latest_abi }}
+      skip_build: ${{ steps.check-abi-ahead.outputs.skip_build }}
     steps:
       - name: Check out code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          sparse-checkout: ""
+          sparse-checkout: "include/gpac/version.h"
 
       - name: Get latest `abi-*` tags
         id: get-tags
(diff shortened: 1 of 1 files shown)


aureliendavid:
Hi, 

Looks good, thanks. 

I have one question about this workflow: 

if we change the ABI manually in version.h (when doing a release for example), how will the workflow handle that? will it re-change or re-commit something? 

if we arbitrarily decide tomorrow that the current state of the code should be ABI 16.2 I think the script should just accept that and maybe move the tags and nothing else?

any opinion on this?

DenizUgur:
Good question. As long as you commit the ABI change with the appropriate tags `abi-16` and `abi-16.2` then the workflow would see that last commit is same as the last abi change and skip it. See the conditional on the build job.

DenizUgur:
And FYI, the push commit here wouldn't trigger other GitHub actions right now. We would need a Deploy Key to trigger other push actions (if that's something we need).

aureliendavid:
> Good question. As long as you commit the ABI change with the appropriate tags `abi-16` and `abi-16.2` then the workflow would see that last commit is same as the last abi change and skip it. See the conditional on the build job.

Ok so we would still have to move the tags manually? 

Couldn't we make it so that the workflow detects that the abi read from version.h is superior to the current ref and in this case just move the tag and do nothing else?

I'm asking because I recently wrote a [release script](https://github.com/gpac/gpac/blob/master/do_release.py) that allows to change the abi version along with the new release tag ([--abi](https://github.com/gpac/gpac/blob/master/do_release.py#L191)). 

For a release we want to push the release tag and the associated commit in one go with something like `git push --atomic origin master v2.6`.

One fundamental thing about this is that the buildbot has to build this exact commit, because we detect releases in scripts by looking for a `-rev0` in the revision. 

If a workflow snicks a commit in after the release commit+tag and before the buildbot starts, this new commit will be built instead and it will have a `-rev1` index and the rest of the release process won't work automatically. 

I assume that since we now have a separated process to handle abi changes that we probably won't have to change the abi version with a new release so this wouldn't be a problem. But we still might for one reason or another. 

Any opinion on this?

DenizUgur:
Just to clarify something, what do you mean by "moving the tags"? We don't touch existing abi tags but only add new ones if it's necessary.

For what you want, I could check if the existing abi is superior to the latest tag. In that case, like you said, I could just tag the commit that triggered the workflow with the appropriate tags.

aureliendavid:
> Just to clarify something, what do you mean by "moving the tags"? We don't touch existing abi tags but only add new ones if it's necessary.
> 

yes you're right, I meant create the new tags for the new abi version

> For what you want, I could check if the existing abi is superior to the latest tag. In that case, like you said, I could just tag the commit that triggered the workflow with the appropriate tags.

if that's doable, that'd be perfect for my scenario I think!

DenizUgur:
In any case, it's better to tag abi tags from the release script along with the version tag.

I'll still add the checks though.

DenizUgur:
I've updated the workflow, can you check it with your release script and see if it behaves as you desired?

aureliendavid:
Thanks, I tested it on a fork and it seems to be working well. 

This was the sequence I tested: 

1. merge the branch

leads to the abi workflow running and moving the abi to 14.0 with this commit: https://github.com/aureliendavid/gpac/commit/956213114e132d5c1ee81a3b3d1900ed5619f1a0

2. do a release 

with `./do_release.py --abi 16.2 2.6.2`

the workflow runs, sets the abi-* tags and quits: https://github.com/aureliendavid/gpac/actions/runs/19961905782/job/57244105762

3. do a non abi related commit

the workflow runs, detects no changes, and stops: https://github.com/aureliendavid/gpac/actions/runs/19961997218

4. do an abi changing commit

the workflow correctly starts back from the abi tag from the release and commits the news changes: https://github.com/aureliendavid/gpac/commit/43d923c1902c16ea4895c1dab402b49f894d613e

This seems perfect to me. 

----

With that sorted I have just one additional concern (sorry for nitpicking). 

The script currently bumps the major value if the abis are deemed incompatible. If I'm not mistaken deleting one abi function is enough to make two versions incompatible right?

This means we could have a scenario where one dev implements a change that refactors/deletes some abi functions. The feature is merged and the abi major is bumped. The next day we realize the feature is incomplete and some other function could also be changed or deleted. The abi major bumps again with the fix. And so on. 

This might be a bit far fetched but I feel like the version could explode pretty quickly with this?  But maybe I'm being too cautious. 


In any case thanks for the changes!

DenizUgur:
In my opinion ABI version is only there to signal whether an SO is compatible across commits or not. So even if we have ABI 100.0 next week it shouldn't really mean anything other than we can't use the same SO between today and next week.

But you're right to be concerned about it, even if one of us works on the master branch and pushes several commits that change the ABI it could bump the ABI more frequently then we really need.

One thing we could do is that since we now have nightly releases from GitHub Actions we could check ABI changes before that, so that we limit the version bump to once a day. Not sure how other projects approach this though.

aureliendavid:
> One thing we could do is that since we now have nightly releases from GitHub Actions we could check ABI changes before that, so that we limit the version bump to once a day. Not sure how other projects approach this though.

I like the idea of doing it at most once a day. (It could also have less chances of conflicts when there are pushes just before or after the workflow is also pushing to the repo)

I can merge this as is for now and if it causes some issues we can revisit the nightly idea. 

Thanks

PR #3405, opened 2025-12-04
ttml: fix negative rewritten timestamps

rbouqueau:
@jeanlf : must have been present for a long time, but was discovered because of #3399.


jeanlf:
thanks!

PR #3410, opened 2025-12-10
add mabr gateway stats

nlsdvl:
adds a /stats endpoint to mabr gateway counting requests served from multicast cache vs unicast


rbouqueau:
Thank you!

PR #3414, opened 2025-12-21
Usability: set http out dir defaults

rbouqueau:
This has become more annoying with GPAC integrations - typically other frameworks or languages. Logs are not always easily accessible and this is a major source or avoidable issues.


jeanlf:
thanks!

PR #3415, opened 2025-12-21
sched: select "lock" by default on select ARM targets

rbouqueau:
Some armv8 64 bits platforms shows rare random issues with our lockfree queue. The memory model is probably broken somehow on this arch and this is not easy to fix reliably. This PR is a workaround until we take a decision about lockfree queues in GPAC to support all archs.


jeanlf:
OK to add this for the time being until we figure out what's wrong with free on these platforms

PR #3421, opened 2026-01-07
mp4box -dxml: consider uncyphered subtype

rbouqueau:
Needed to be iso with `gpac -i inspect:deep:analyze=bs` e.g. with OBUs (which are currently dumped as generic NHML streams).


jeanlf:
thanks !

PR #3424, opened 2026-01-10
Update FreeBSD support

ocochard:
Fix linker flags, path detection, and /proc handling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ endif
 	$(INSTALL) $(INSTFLAGS) -m 644 $(SRC_PATH)/share/gui/gwlib.js "$(DESTDIR)$(prefix)/share/gpac/gui/"
 
 
-ifeq ($(CONFIG_DARWIN),yes)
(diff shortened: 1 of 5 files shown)


rbouqueau:
Thank @ocochard for your contribution 🙏 . @jeanlf @aureliendavid Ready to go. However don't we want to add a FreeBSD flavor to our buildbot infra ?

aureliendavid:
Yes sorry I haven't had time to do it yet but I'll see if we can add some github actions to test the build on more systems, like freebsd. If you have others in mind I can try to add them too. (probably won't do a full buildbot worker for those though). 

If I can't do it next week I'll just merge the PR and add the builder later. 

in any case, thanks for the contrib!

aureliendavid:
I merged the PR and added a github action to check the build of gpac for FreeBSD (using anyvm.org) on pushes: see example run at  https://github.com/gpac/gpac/actions/runs/21180333988/job/60920601035#step:3:3280

thanks!

PR #3439, opened 2026-01-27
eAC-3: remove test-mode conditional

rbouqueau:
partially reverts f3d2a1a55868


rbouqueau:
I'm going to open a parallel PR re-using the existing `ase` parameter.

rbouqueau:
The thing with `asemode` is that it sets the default at `v0-bs` which sets the values before Dolby corrected them. We have reports saying the old value would offer a better compatibility beside being non compliant with ETSI TS 102 366 V1.4.1.

rbouqueau:
Superseded by #3443

PR #3447, opened 2026-02-03
GPAC MCP Server

DenizUgur:
A very simple MCP server for gpac


jeanlf:
Thanks for this initial work !

PR #3451, opened 2026-02-08
avoid error-level log message being split

rbouqueau:
With strict error, the end of the message would never display.


jeanlf:
Thanks !

PR #3454, opened 2026-02-10
keep data of unknown sample entry types

maryla-uc:
Previously, unknown sample entries were first parsed as unknown boxes (unkn) which correctly store their raw data, but then reparsed and replaced with generic sample entry boxes (gnrm/gnrv/gnra) in a way that dropped the box's data.

Fix missing space when outputting these boxes to xml.

### DIFF
--- a/src/isomedia/box_code_base.c
+++ b/src/isomedia/box_code_base.c
@@ -1583,9 +1583,18 @@ GF_Box *gnrm_box_new()
 	return (GF_Box *)tmp;
 }
 
-//dummy
 GF_Err gnrm_box_read(GF_Box *s, GF_BitStream *bs)
 {
+	GF_GenericSampleEntryBox *ptr = (GF_GenericSampleEntryBox *)s;
+	GF_Err e = gf_isom_base_sample_entry_read((GF_SampleEntryBox *)s, bs);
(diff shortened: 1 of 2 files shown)


maryla-uc:
See a [test file](https://github.com/user-attachments/assets/2e1b2b13-f8f3-4cd7-aa86-4265662f9827)  (modified from [timed-metadata.mp4](https://github.com/MPEGGroup/FileFormatConformance/blob/main/data/file_features/published/isobmff/timed-metadata.mp4), BSD 3-Clause Clear License)
e.g.: `MP4Box -ref 1:test:2 timed-metadata-unknowntype.mp4 -out out.mp4`
would lose the data of the unknown 'TEST' box, but now keeps it.

rbouqueau:
Thanks, looks like a useful thing. I'd like to review this more carefully.

rbouqueau:
LGTM but asking @jeanlf for review.

jeanlf:
Thanks !

aureliendavid:
thanks for the contrib! 

just keep in mind memory leaks when returning right after creating local objects (see e6d3098)

maryla-uc:
Indeed, thanks for the fix!

PR #3504, opened 2026-03-25
fix multi-key write format to use `use_mkey` flag

rafaelcaricio:
`sgpd_write_entry()` checked nb_keys>1 to decide whether to write the multi-key format, but `sgpd_size_entry()` uses use_mkey. When multi-key is enabled with a single key (e.g. `multiKey="subs=1,0"` for per-subsample encryption), the write path skipped the 2-byte `key_count` field, causing a box size mismatch and corrupted output.

This change uses the same `use_mkey` flag that `sgpd_size_entry()` already uses.


jeanlf:
Indeed, thanks for the fix !

PR #3525, opened 2026-04-10
dasher live with stl: don't display new period if not segment timeline entry

rbouqueau:
=> to avoid wrong $Time$ (instead of PTO) resolving.

### DIFF
--- a/src/media_tools/mpd.c
+++ b/src/media_tools/mpd.c
@@ -4657,6 +4657,57 @@ GF_Err gf_mpd_write_m3u8_master_playlist(GF_MPD const * const mpd, FILE *out, co
 	return GF_OK;
 }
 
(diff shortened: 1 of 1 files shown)


rbouqueau:
@jeanlf The [dash_merge_mperiod](https://tests.gpac.io/test/dash_merge_mperiod/buildbot-dash-period-stl-empty) test exactly does what the patch checks. Should I conclude that the serializer is not the right place to fix this, and we need to hold the `Period`creation in the `dasher`?

jeanlf:
Can you check if [this commit](https://github.com/gpac/gpac/commit/4bfbf765e993ab97fc6341e0438d12cf92223b40) fixes your issue ?

rbouqueau:
@jeanlf Your commit is perfect. Should we close here and merge yours?

jeanlf:
will use #3536 instead

Issue #2828, opened 2024-05-06
SCTE-35 manifest support


rbouqueau:
Meant to be executed after https://github.com/gpac/gpac/pull/2748 is merged.

We still need to add the manifest manipulation:
- [ ] DASH: SCTE-35: add ```<InbandEventStream schemeIdUri="urn:scte:scte35:2013:bin"></InbandEventStream>``` => DASH-IF recommends xml+bin
- [ ] HLS: Add ```#EXT-X-CUE-IN``` and ```#EXT-X-CUE-OUT```.

jeanlf:
where are we on this one ? Can we close ?

rbouqueau:
Still not done. Let me keep the tab open to see if this can be processed in the upcoming weeks. Would be a nice addition.

rbouqueau:
Draft PR: #3534.

I have some local updated doc for https://wiki.gpac.io/Howtos/scte35-markers/. Same for tests I need to translate into the test suite.

We're DVB-DASH compliant.

I plan to add HLS very soon.

Note: `evte_agg` only works with Event Tracks i.e. muxed content. Same as for subs where this type of content exists in raw and in isobmff.

rbouqueau:
HLS added on the PR branch.

Issue #3342, opened 2025-09-16
SCTE-35 Preservation in Live MPEG-DASH with dash-ctx


moises-xr:
Hi,

We are seeking assistance with preserving SCTE-35 markers from MPEG-TS segments to MPEG-DASH
manifests in a live streaming scenario. The issue occurs specifically when using the -dash-ctx parameter,
which is crucial for our continuous live streaming workflow.

### **Environment**

- GPAC Version: MP4Box - GPAC version 2.5-DEV-rev1772-g97c35d2a2-master
- Use Case: Live streaming with continuous 4.8-second MPEG-TS segments containing SCTE-35 markers
- Target Output: MPEG-DASH with preserved SCTE-35 signaling

### **Current Workflow**

**Step 1: Live Stream Segmentation with FFmpeg**

We capture RTSP streams and create MPEG-TS segments using FFmpeg:

```shell
./ffmpeg-6.1.1/ffmpeg -y -hwaccel cuda -hwaccel_output_format cuda \
-i $rtsp_url \
-map 0:v -vcodec hevc_nvenc -r 60 -b:v 50000k -preset fast -fps_mode cfr
\
-copyts -muxdelay 0 -muxpreload 0 \
-f segment -g 288 -segment_start_number 1 -segment_time 4.8 \
/tmp/segments/video/output_%05d_vFLAT-1.ts \
-map 0:a -af aresample=async=1 -ar 48000 -acodec aac -b:a 256k \
-f segment -segment_start_number 1 -segment_time 4.8 \
/tmp/segments/audio/output_%05d_1000-256.ts
``` 

**Step 2: TS to MP4 Conversion**

```shell
MP4Box -add /tmp/_0-8000/tmp_input_0-8000.ts /tmp/_0-8000/tmp_input_0-
8000.mp4
``` 

**Step 3: MP4 to MPEG-DASH (Current Production)**

```shell
MP4Box -dash-ctx /tmp/_0-8000/tmp_dash_0-8000.txt -dash 600 -dash-profile
live \
-segment-ext m4s -out /tmp/_0-8000/tmp.mpd \
-segment-name livestream_0-8000_ /tmp/_0-8000/tmp_input_0-8000.mp4
``` 

### SCTE-35 Integration Attempts

**Test Input Analysis**
Our test file contains SCTE-35 data:

```shell
ffprobe scte35_spliceInsert_2hour_demo.ts
[...]

PR #3534, opened 2026-04-16
scte35 out of band (in manifest)

rbouqueau:
The goal is to address #2828. I'm look for some feedback before I upload the tests and wiki update I also have:
- DASH: we only support `schemeIdUri="urn:scte:scte35:2013:xml+bin"` as per DASH-IF recommendations and found samples. When looking at streams it seems to be the defacto standard.
- HLS: I plan to add it too. More complex as people use [4 schemes](https://developers.broadpeak.io/docs/dai-scte35#scte-35-signalling-in-hls-playlists).

### DIFF
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -181,6 +181,14 @@ enum
 	AC4_OBJECT_BASED_AJOC_CONTENT,
 };
 
+// SCTE-35 HLS signalling
+GF_OPT_ENUM (Scte35ModeHLS,
+    SCTE35_DATERANGE,
+    SCTE35_HLS_OAT_CLS,
+    SCTE35_SPLICEPOINT,
+    SCTE35_CUE,
+);
+
 //these are not exported for now
 //get destination name by index
 char *gf_filter_pid_get_destination_ex(GF_FilterPid *pid, u32 dst_idx);
@@ -248,6 +256,7 @@ typedef struct
 	DasherWaitLastPktCtrl seg_sync;
 	Bool hls_ap;
 	Scte35Mode scte35;
+	//Scte35ModeHLS scte35_hls;
 
 	//internal
 	Bool in_error;
@@ -12087,6 +12096,7 @@ static const GF_FilterArgs DasherArgs[] =
 	{ OFFS(ttml_agg), "force aggregation of TTML samples of a DASH segment into a single sample", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_EXPERT},
 	{ OFFS(evte_agg), "force aggregation of Event Track samples of a DASH segment into a single sample", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_EXPERT},
 	SCTE35_ARG,
+	//{ OFFS(scte35_hls), "HLS SCTE-35 signalling", GF_PROP_UINT, "cue", "daterange|oatcls|splicepoint|cue", GF_FS_ARG_HINT_EXPERT },
 
 	{ OFFS(force_flush), "deprecated - use sflush instead", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_HIDE},
 	{ OFFS(base64), "embed init segments in manifests as base64", GF_PROP_BOOL, "false", NULL, GF_FS_ARG_HINT_ADVANCED},
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -151,6 +151,14 @@ GF_OPT_ENUM (DasherWaitLastPktCtrl,
 	DASHER_SEGSYNC_AUTO,
 );
 
+// DASH SCTE-35 (out-of-band)
+GF_OPT_ENUM (DasherDashScte35Mode,
+	//not implemented: DASHER_SCTE35_DASH_XML,
+	DASHER_SCTE35_DASH_XML_BIN,
+	DASHER_SCTE35_DASH_EVTE,
+	DASHER_SCTE35_DASH_NONE,
+);
+
 // Index mode as used from a GHI (GPAC HTTP Streaming index) demuxer
 enum
 {
@@ -246,6 +254,7 @@ typedef struct
 	GF_DashAbsoluteURLMode hls_absu;
 	DasherWaitLastPktCtrl seg_sync;
 	Bool hls_ap;
+	DasherDashScte35Mode scte35;
 
 	//internal
 	Bool in_error;
@@ -4004,9 +4013,19 @@ static void dasher_open_pid(GF_Filter *filter, GF_DasherCtx *ctx, GF_DashStream 
 		gf_filter_set_source(ds->dst_filter, ttml_agg, szSRC);
 	}
 
-	//inject scte35dec filter
-	if (ctx->evte_agg && (ds->codec_id==GF_CODECID_SCTE35 || ds->codec_id==GF_CODECID_EVTE))
-		dasher_inject_scte35_processor(filter, ds, szSRC);
+	// SCTE-35
+	if (ctx->evte_agg && (ds->codec_id==GF_CODECID_SCTE35 || ds->codec_id==GF_CODECID_EVTE)) {
+		// DASH SCTE-35
+		if (ctx->do_m3u8)
+			GF_LOG(GF_LOG_WARNING, GF_LOG_DASH, ("[Dasher] evte_agg option used with HLS is likely not supported by your player\n"));
+
+		if (ctx->scte35 == DASHER_SCTE35_DASH_EVTE) {
+			//inject scte35dec filter
+			dasher_inject_scte35_processor(filter, ds, szSRC);
+		} else {
+			GF_LOG(GF_LOG_WARNING, GF_LOG_DASH, ("[Dasher] evte_agg option requires the use of event tracks (:scte35=evte): ignoring\n"));
+		}
+	}
 }
 
 static void dasher_set_content_components(GF_DashStream *ds)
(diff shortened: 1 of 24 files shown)


v-faeez-kadiri:
[0001-mp4mux-dasher-add-SCTE-35-in-band-emsg-support.patch](https://github.com/user-attachments/files/27020311/0001-mp4mux-dasher-add-SCTE-35-in-band-emsg-support.patch)

Hi @rbouqueau,

Thank you for the work on out-of-band SCTE-35 EventStream signaling in this PR, it fills an important gap for manifest-level ad marker visibility.

We have a production use case (the same one described in #3342) that requires **in-band** SCTE-35 signaling: emsg boxes embedded directly in fMP4 segments. This is the complementary signaling mode to what this PR provides, and both are recommended by DASH-IF IOP Section 4.7 for full interoperability.

### Why in-band is needed

In broadcast / live TS-to-DASH workflows, SCTE-35 markers arrive as splice_info_sections in the transport stream and need to be carried through to the DASH output as emsg boxes with scheme `urn:scte:scte35:2013:bin`. This is the standard signaling path for client-side ad insertion (CSAI) where DASH players (dash.js, Shaka, ExoPlayer) listen for inband events to trigger ad decisions in real-time. Out-of-band EventStream (`xml+bin`) serves server-side and manifest-aware workflows, but does not cover this case.

The specific issue we hit was that `dash-ctx` mode (required for continuous live segmentation) stalled when evte tracks were present, because the dasher created output PIDs for them but the muxer had no consumer. Our patch resolves this and adds the full emsg emission pipeline.

### What the patch does

We have a single squashed patch (attached below, applies cleanly on current master `1a1103b3b`) that adds:

- **emsg generation in mp4mux**: three emission paths, direct from TS SCTE-35 packet properties, pending event queue for segment-aligned emission, and dash-ctx mode reading evte tracks from source MP4 files
- **InbandEventStream in dasher**: automatically adds `<InbandEventStream schemeIdUri="urn:scte:scte35:2013:bin"/>` to MPD AdaptationSets when SCTE-35/EVTE streams are detected
- **Shared SCTE-35 parser**: `scte35dec_get_timing()` extracted to `include/gpac/internal/scte35.h` so both `dec_scte35` and `mux_isom` use the same parsing code (no duplication)
- **33-bit PTS wraparound handling**: `(splice_time + pts_adjustment) & 0x1FFFFFFFF` applied to both `splice_insert` and `time_signal` paths per SCTE-35 2022 Section 9.2
- **Unit tests**: three new test vectors covering `time_signal`, `splice_insert`, and PTS wraparound
- **Bug fixes**: `return` vs `continue` in dmx_m2ts SCTE-35 PID loop, NULL value field crash in mpd.c InbandEventStream serialization, descriptor component skip (48 bits = 6 bytes, not 48 bytes)

### Coexistence with out-of-band

The two signaling modes are cleanly separated:

| Aspect | This PR (out-of-band) | Our patch (in-band) |
|--------|----------------------|---------------------|
| Scheme URI | `urn:scte:scte35:2013:xml+bin` | `urn:scte:scte35:2013:bin` |
| MPD element | `<EventStream>` | `<InbandEventStream>` |
| Carrier | MPD XML | emsg box in segments |
| Use case | SSAI / manifest-aware | CSAI / player-driven |

Both can coexist in the same MPD and the same DASH session without conflict. A deployment can enable either or both depending on whether ad decisions happen server-side, client-side, or both.

One area worth aligning on: our patch discards evte PIDs in the dasher (since the muxer reads them directly from the source MP4 in dash-ctx mode), while your `evte_agg` option re-segments them. If both modes are eventually merged, we may want a configuration flag to select the evte handling strategy.

### Patch

The patch is generated with `git format-patch` against current master.

We would be happy to adjust the patch based on your feedback, split it differently, or adapt it to align with the direction you prefer for SCTE-35 support overall.

Thanks for considering this.

rbouqueau:
Hi, thanks for this contribution. I'm working on it. I've already pushed 3 sub-commits based on your work.

The overall patch is quite complex. I'd like to test what you cover and add it to our regression testsuite. Do you have corresponding command-lines for your use-cases?

I knew I had to signal the evte -> mpd (inband) next. Planning to work on this today.

v-faeez-kadiri:
Hi @rbouqueau,

Thanks for picking those up so quickly, much appreciated.

Regarding command-lines: we cover two distinct workflows. The VOD path uses only GPAC tooling and should be straightforward to add to the test suite. The Live path has an external dependency that is worth being upfront about.

### VOD workflow

This one is self-contained. Given a TS source with SCTE-35:

```bash
# create MP4 with evte track
MP4Box -add input.ts#video \
       -add input.ts#video:@scte35dec:segdur=0/0 \
       -new video_with_scte35.mp4

MP4Box -add input.ts#audio -new audio.mp4

# DASH with emsg
MP4Box -dash 2000 -dash-profile live -segment-ext m4s \
       -segment-timeline -url-template -rap \
       -out output/manifest.mpd \
       -segment-name 'seg_$RepresentationID$_' \
       video_with_scte35.mp4 audio.mp4
```

Standard FFmpeg works fine here if transcoding is not needed beforehand, since the full TS file is handed to GPAC intact. The emsg boxes carry the actual `splice_time + pts_adjustment` from the SCTE-35 binary, not the packet arrival time.

### Live workflow (dash-ctx)

This is the #3342 scenario. Segments arrive incrementally from a live encoder, so `dash-ctx` is required for state continuity. The GPAC side looks like this:

```bash
# per-segment: TS -> MP4 with evte track
MP4Box -add "$segment.ts#video" \
       -add "$segment.ts#video:@scte35dec:segdur=0/0" \
       -new "$segment.mp4"

# incremental DASH
MP4Box -dash-ctx dash_ctx.txt \
       -dash 600 -dash-profile live \
       -segment-ext m4s -url-template \
       -out manifest.mpd \
       -segment-name 'seg_$Number$' \
       "$segment.mp4"
```

The complication is upstream of GPAC: standard FFmpeg drops SCTE-35 sections during segmentation. We use a custom FFmpeg build that explicitly maps and copies the SCTE-35 data PID (`-map 0:N -c:d copy` where N is the scte_35 stream index). Without that, the TS segments arriving at GPAC contain no SCTE-35 to extract. We validate with TSDuck (`tsp -P splicemonitor`) between FFmpeg output and GPAC input to confirm markers survive.

This is not a GPAC concern per se, but it does affect reproducibility for the test suite, the Live path cannot be tested end-to-end with a stock FFmpeg build. For the VOD path there is no such constraint.

### For the regression suite

The VOD workflow should be directly usable. For Live, if you have (or can generate) pre-segmented TS files that already contain SCTE-35, the GPAC portion of the pipeline works identically, it is only the segmentation step that requires the custom FFmpeg.

v-faeez-kadiri:
You can use the attached pre-segmented TS files
https://drive.google.com/file/d/1HnSzZD0yIi_xOnC8HMf0u61MGrdhEuHP/view?usp=sharing

rbouqueau:
Thanks. I think I was confused by the mention to `evte_agg` in your initial message: since your content is pre-segmented, it is indeed not necessary.

I also don't understand why you can't just ingest the TS segment to DASH it. Do you know the reason for that?

I made a list of the scenarios we've seen and that I'm trying to resolve/simplify:

Use-case | Input | Output | CL | Status
--- |--- |--- |--- |---
VoD | pre-packaged | oob evte track | `dasher` | [this PR](#3534)
VoD+live | pre-packaged or live `evtapi`-style | oob evte track | `dasher` | [this PR](#3534)
VoD+Live | TS(+MP4? cf my question in this message) | inband `emsg` | `-dash-ctx` | [patch](https://github.com/gpac/gpac/pull/3534#issuecomment-4313274507)
Live+Nielsen | TS | TS or `dasher` | id3 + manual `:inband_event=` | [this PR](#3534)

PS: table post-edited

@v-faeez-kadiri Does that look correct to you ?

v-faeez-kadiri:
Hi @rbouqueau,

The intermediate MP4 step exists for two reasons: 

1. It extracts the actual splice_time + pts_adjustment from the SCTE-35 binary for accurate emsg presentation_time, DASHing the TS directly would use packet CTS which is 4–7s early, 
2. Our production architecture packages video, audio, and subtitle TS segments independently on separate servers, the TS-to-MP4 step isolates SCTE-35 extraction into the video path.

Your table looks correct. I would add one nuance: the VOD path from our patch also produces inband emsg (not just your oob EventStream), so both can coexist for the same input. And the Live row specifically requires the TS - MP4-with-evte conversion step for the timing reason above.

rbouqueau:
Thanks, I've updated the table accordingly.

Note that you could merge the two steps by replacing `MP4Box` by the `gpac`tool.

I need to think more about how to resync the SCTE-35 data to re-inject the emsg boxes.

src/filters/dasher.c
```diff
@@ -28,6 +28,7 @@
 #include <gpac/iso639.h>
 #include <gpac/mpd.h>
 #include <gpac/internal/media_dev.h>
+#include <gpac/internal/scte35.h>
```
aureliendavid, line 31: maybe a missing git add / push there?

rbouqueau: you're right, thanks for the msg, I'm fixing it

aureliendavid:
another testing/buildbot quirk: there exists a testsuite branch of the same name: https://github.com/gpac/testsuite/tree/buildbot-scte35-manifest so the PR is tested against that one, but it looks pretty old and causes issues (due in part to the -old-arch refactor)

it should probably be rebased (or merged, or deleted) to allow testing the gpac branch on something more recent

rbouqueau:
@v-faeez-kadiri I could find some time to explore your patch. It was squashed, probably a PR with multiple commits would have been easier to process for me. Thanks anyway for the effort and chat here - highly appreciated.

The TS->DASH in-band pipeline works in my draft PR in the same way as your patch. It was very similar to the id3 processing already so I took the liberty to simplify your code.

The EVTE/MP4->DASH in-band doesn't. We don't like the design of opening the track from the isom muxer. The best way to re-inband the content is to remux (e.g. to mpeg-ts). At the same time, while in-band with `emsg` boxes exist, the media industry seems to have moved 5+ years ago to `evte`track or in-band in the manifest ([dash-if](https://dashif.org/docs/IOP-Guidelines/DASH-IF-IOP-Part5-v5.0.0.pdf)) or "[is not implemented in all of the main browser engines](https://www.w3.org/TR/media-timed-events/#mpeg-dash-and-iso-bmff-emsg-events)". I was thinking to extend scte35dec to attach export the "scte35" property from an `evte track` but that would only solve half of the problem as it would require to implement a mux anyway (i.e. we would need to ensure the `evte` samples are dispatched before the other media it needs to attach too).

Any opinion appreciated here. I still need to fix some DASH signalling, add HLS, and add tests to the regression testsuite.

rbouqueau:
@aureliendavid Any idea about the Ubuntu 32+64 build failures ? I can't reproduce on my computer (on Debian) + Ubuntu 64 VM. Could I easily work from the same VMs ?

aureliendavid:
The check fail links to the buildbot logs: https://buildbot.gpac.io//#/builders/12/builds/2461

it show that it fails when building the unittests

it complain about the ceil function during link so I guess a `-lm` missing

reproduced with: 

```sh
docker run -it gpac/ubuntu bash
rm .git/config
git remote add origin https://github.com/gpac/gpac.git
git fetch && git checkout buildbot-scte35-manifest
make distclean ; ./configure --enable-debug --unittests ; make -j unit_tests
```

with this patch: 

```diff
diff --git a/src/Makefile b/src/Makefile
index c94c2ad06..b2ab0a0b0 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -429,7 +429,7 @@ ifeq ($(UNIT_TESTS),yes)
                -o ../bin/gcc/unittests \
                ../bin/gcc/unittests.c \
                $(shell find $(SRC_PATH) -path "*/unittests/*.c" | grep -v bin | sort) \
-               -Wl,-rpath,$(realpath ../bin/gcc) -L../bin/gcc -lgpac
+               -Wl,-rpath,$(realpath ../bin/gcc) -lm -L../bin/gcc -lgpac
 endif

```

it builds

rbouqueau:
@jeanlf Ready to merge. See also https://github.com/gpac/testsuite/pull/57 and https://github.com/gpac/wiki/pull/49.

I still need to stack some other changes for the isom/evte -> dash outband EventStream config. But this can wait for a next PR if I'm not fast enough. This one is big enough already ;)

jeanlf:
LGTM, I let you merge

PR #3536, opened 2026-04-17
do not serialize periods if empty SegmentTimeline with $Time adressing - r…

jeanlf:
…ework of #3525

PR #3541, opened 2026-04-20
mpd: don't serialize when no unskipped period - 

rbouqueau:
follows up 4bfbf765e993a and https://github.com/gpac/gpac/pull/3536#issuecomment-4280437335


jeanlf:
thanks !

PR #3543, opened 2026-04-22
GPAC_QUIC_METRICS: Add unified metrics for HTTP/1.1, HTTP/2, HTTP/3

Husnkhalil:
This PR introduces unified metrics collection across HTTP/1.1, HTTP/2, and HTTP/3. in addition to a tunning option on the server side.. 
These Metrics includes:
- Time to First Byte (TTFB)
- Total transfer duration
- Throughput (Mb/s)
- CPU usage (user/sys)
- Memory usage (RSS)
- 
Implementation is isolated to downloader modules.
Tested locally on H1, H2, and H3.

### DIFF
--- a/src/utils/downloader.c
+++ b/src/utils/downloader.c
@@ -22,9 +22,9 @@
  *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
  *
  */
-
+#include <inttypes.h>
 #include "downloader.h"
-
+#include <gpac/tools.h>
 #ifndef GPAC_DISABLE_NETWORK
 
 static void gf_dm_connect(GF_DownloadSession *sess);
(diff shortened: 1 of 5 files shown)


aureliendavid:
Hi, 

Thanks for the contrib. 

Just a quick issue: 

you're using `sys/resource.h` and `rusage` directly in downloader.c - these are posix specific and won't build on windows (further there's a macro to prevent building this on windows but no alternative for windows builds)

windows has stuff like `GetProcessTimes()` and `GetProcessMemoryInfo()`

also it looks like `ru_maxrss` can have different units on linux (KB) and macos (bytes): https://stackoverflow.com/questions/59913657/strange-values-of-get-rusage-maxrss-on-macos-and-linux (you have a comment about this so I think you're aware of it)

my advice would be to externalize all the platform specific code with all the ifdefs to utility functions (in os_divers.c for example), then you can export them and use these portable functions in the downloader modules for better readability 

you can check what gf_sys_get_rti() is doing and how it's defined (you can probably even re-use it directly to get process info)

Husnkhalil:
hello,
Thanks for the feedback!

I removed direct usage of getrusage and replaced it with gf_sys_get_rti() across H1/H2/H3 to ensure portability and consistent units.
Metrics are now unified and platform-specific code has been removed from the downloader modules.
thank you again!!
Let me know if adjustments needed.

jeanlf:
thanks, we only added h2 and h3 options but won't include the metrics part for now - we'll cleanup the log later

PR #3567, opened 2026-05-12
Support extracting standalone IAMF from MP4.

jwcullen:
Support extracting standalone IAMF from MP4.

  - Enables extracting IAMF tracks for IAMF containers into "standalone" IAMF streams.
  - Works via any standard GPAC pileine commands (e.g. `gpac -i input.mp4 -o otuput.iamf`)
  - `descriptors.c`: Split out a function, standalone IAMF uses a subset of the IACB (`config_OBUs`).
  - `write_generic.c`: IAMF has key frames, so bypass SAP check.

### DIFF
--- a/include/gpac/mpeg4_odf.h
+++ b/include/gpac/mpeg4_odf.h
@@ -1970,7 +1970,7 @@ typedef struct
 */
 GF_IAConfig *gf_odf_iamf_cfg_new();
 
-/*! writes IAMF config to buffer
+/*! writes IAMF config (IACB) to buffer
 \param cfg the IAMF config to write
 \param outData set to an allocated encoded buffer - it is the caller responsibility to free this
 \param outSize set to the encoded dsi buffer size
@@ -1978,12 +1978,19 @@ GF_IAConfig *gf_odf_iamf_cfg_new();
 */
 GF_Err gf_odf_iamf_cfg_write(GF_IAConfig *cfg, u8 **outData, u32 *outSize);
 
-/*! Writes the IAMF config to bitstream
+/*! Writes the IAMF config (IACB) to bitstream
 \param cfg the IAMF config to write
 \param bs the bitstream object
 \return error code if any
 */
 GF_Err gf_odf_iamf_cfg_write_bs(GF_IAConfig *cfg, GF_BitStream *bs);
+
+/*! Writes the IAMF config payload (OBUs only) to bitstream
+\param cfg the IAMF config to write
+\param bs the bitstream object
+\return error code if any
+*/
+GF_Err gf_odf_iamf_cfg_write_obus(GF_IAConfig *cfg, GF_BitStream *bs);
 
 /*! IAMF config destructor
 \param cfg the IAMF config to destroy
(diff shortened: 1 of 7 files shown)


jwcullen:
Depends on test suite updates in GPAC/testsuite#59. Please see the diff there. This PR results in better information reporting when inspecting IAMF.

Please let me know if I need to structure this PR differently to reference the updated test suite.

rbouqueau:
Thanks for your contribution. The code LGTM. Can you fix the 'testsuite' submodule conflict by not updating its revision in the PR?

Also can you elaborate on why the two hashes from another IAMF test in the testsuite changed? Was that part of what you wanted to improve?

jwcullen:
I updated this PR to not touch the test suite submodule.

So in MP4, IAMF is configured with the `iacb` box [IAMF §6.2.4](https://aomediacodec.github.io/iamf/#iaconfigurationbox-section).

```
  unsigned int (8) configurationVersion = 1;
  leb128() configOBUs_size;
  unsigned int (8 x configOBUs_size) configOBUs;
```

In standalone IAMF, the first two fields are not present. So typically a standalone IAMF bitstream with start with an IA Sequence Header (0xf8 = decimal 248).

When converting from MP4 -> standalone IAMF, those extra fields need to be handled consistently and dropped if present. As part of cleaning this up, I realized some places in GPAC were mixed on whether they were just storing the `configOBUs` or the whole `iacb` box.

```
- [IAMF] Unknown configurationVersion 248
```

The original output file (before gpac/testsuite#59) shows language like "Unknown configurationVersion 248", because this was a section where the whole `iacb` was expected, but in fact the first bytes were the bare IA Sequence Header.

rbouqueau:
@jwcullen Thanks. LGTM. Waiting for a second review before merging.

jwcullen:
IAMF has a technical detail where standalone IAMF is allowed to (but not required) to have Temporal Delimiter OBUs under (Simple, Base, Base-Enhanced) profiles [§2.3.3.2](https://aomediacodec.github.io/iamf/#iadata). But these OBUs are forbidden in MP4 [§6.2.5](https://aomediacodec.github.io/iamf/#iasampleformat). Since it's never required, this implementation leaves them absent when extracting.

I also tested the change locally on all of the relevant (Simple, Base, Base-Enhanced) profile `libiamf` test vectors (https://github.com/AOMediaCodec/libiamf/tree/main/tests). These provide two flavors of MP4, and standalone IAMF.

The ones that match all of these conditions are byte-by-byte identical:

  - Represent valid IAMF.
  - Agree with the profiles GPAC supports.
  - Exclude temporal delimiters (most of them)

I'm just bringing this up, because in future profiles, there might need to be further work around these temporal delimiters. But this PR is complete for (Simple, Base, Base-Enhanced).

aureliendavid:
hi, 

see comments in https://github.com/gpac/testsuite/pull/59

once fixed you can open a new testsuite PR and **set the base branch to [mp4_to_iamf_filters](https://github.com/gpac/testsuite/tree/mp4_to_iamf_filters) instead of master** to properly run our tests against it

thanks

aureliendavid:
tests updated in https://github.com/gpac/testsuite/pull/61 

/buildbot-rebuild

aureliendavid:
tests now look ok, if ok to merge please merge https://github.com/gpac/testsuite/pull/60 first

rbouqueau:
Many thanks!

Issue #3588, opened 2026-05-28
GPAC has an uncontrolled decompression vulnerability in gf_gz_decompress_payload_ex


HMF2021:
# GPAC has an uncontrolled decompression vulnerability in gf_gz_decompress_payload_ex

## Supplier

GPAC is an open-source multimedia framework.

- Website: https://gpac.io/
- Source code: https://github.com/gpac/gpac

## Affected Versions

- GPAC 26.02.0: affected
- GPAC master: affected at the time of testing

The vulnerability is present in both the current GPAC 26.02.0 release and the current master branch.

## Vulnerability File

```text
src/utils/base_encoding.c
src/isomedia/box_funcs.c
```

## Vulnerability Description

GPAC contains an uncontrolled decompression issue in `gf_gz_decompress_payload_ex`. The function inflates attacker-controlled compressed data into a heap buffer that starts at 4096 bytes and is repeatedly doubled with `gf_realloc` until the zlib stream ends or the buffer approaches `GF_UINT_MAX`. The API does not accept a maximum output size and does not enforce a compression-ratio limit.

In the ISOBMFF parser, compressed top-level boxes such as `!mov`, `!mof`, `!six`, and `!ssx` are automatically decompressed before the decompressed payload is parsed as the corresponding normal box. An attacker can therefore provide a small MP4 file containing a compressed box that expands to a very large in-memory payload, causing excessive memory consumption and denial of service.

The vulnerable decompression loop is:

https://github.com/gpac/gpac/blob/v26.02.0/src/utils/base_encoding.c#L252

https://github.com/gpac/gpac/blob/v26.02.0/src/utils/base_encoding.c#L283

```c
while (d_stream.total_in < data_len) {
			err = inflate(&d_stream, Z_NO_FLUSH);
			if (err < Z_OK || err == Z_NEED_DICT) {
				e = GF_NON_COMPLIANT_BITSTREAM;
				break;
			}
			if (err==Z_STREAM_END) break;

			if (size >= GF_UINT_MAX/2 - 1)
				size = GF_UINT_MAX - 1;
			else
				size *= 2;
			*uncompressed_data = (char*)gf_realloc(*uncompressed_data, sizeof(char)*(size+1));
			if (!*uncompressed_data) return GF_OUT_OF_MEM;
			d_stream.avail_out = (u32) (size - d_stream.total_out);
			d_stream.next_out = (Bytef*) ( *uncompressed_data + d_stream.total_out);
		}
```

[...]

PR #3589, opened 2026-05-28
zlib: avoid zip bombs with ZLIB_INFLATE_MAX_MULTIPLIER 

aureliendavid:
Hi all, 

see #3588

our inflate() loops are vulnerable to zip bombs (which are more an annoyance than an actual threat but still)

since we can't know in advance the decompressed output buffer size, the easiest solution is either to have a hard size limit or a multiplier limit

I implemented the multiplier limit in this branch, with a (kinda randomly picked) value of 32 (allowing I think all legitimate uses and preventing really aggressive malicious inputs). 

I'm doing this as a PR in case anyone has comments on the solution or the value of ZLIB_INFLATE_MAX_MULTIPLIER 

In case of no comments I'll merge this (on monday probably).

cheers

### DIFF
--- a/include/gpac/setup.h
+++ b/include/gpac/setup.h
@@ -916,6 +916,12 @@ size_t gf_strlcat(char *dst, const char *src, size_t dsize);
 #endif
 //! @endcond
 
+#ifndef GPAC_DISABLE_ZLIB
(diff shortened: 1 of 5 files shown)


jeanlf:
LGTM

rbouqueau:
LGTM too, we let you @aureliendavid merge

PR #3593, opened 2026-06-01
outhttp: prevent root escaping by checking the realpath() of resources

aureliendavid:
like we talked about, this adds a portable version of the posix realpath() function and uses it to check the paths of file requests to the webserver, insuring the resolved path is a child of the root dir

comments welcomed, will merge soon otherwise 

thanks


jeanlf:
LGTM

PR #3594, opened 2026-06-01
dasher: add option to switch Period when new encryption key is encountered

rbouqueau:


src/filters/dasher.c
```diff
@@ -1816,6 +1820,28 @@ static GF_Err dasher_configure_pid(GF_Filter *filter, GF_FilterPid *pid, Bool is
...
+		u32 cenc_key_info_crc = 0;
+
+		p = gf_filter_pid_get_property(pid, GF_PROP_PID_CENC_KEY_INFO);
+		if (p && (p->type==GF_PROP_DATA) && p->value.data.ptr && p->value.data.size) {
+			has_cenc_key_info = GF_TRUE;
+			cenc_key_info_crc = gf_crc_32(p->value.data.ptr, p->value.data.size);
+		}
+
+		if ( is_reconfigure && ds->cenc_key_info_init && 
+			((has_cenc_key_info != ds->has_cenc_key_info) || (has_cenc_key_info && (cenc_key_info_crc != ds->cenc_key_info_crc))) ) {
```
jeanlf: Not sure we need ds->has_cenc_key_info, the CRC should be enough no ?

rbouqueau: indeed thx: adding a testsuite branch then amending to trigger build

PR #3669, opened 2026-06-25
iamf: set SAP 1 by default when flushing IAMF samples 

felicialim:
- Default `GF_FILTER_SAP_1` when flushing IAMF samples (always a sync sample).
- Avoid relying on uninitialized AV1 video keyframe state, which wrote an incorrect `stss` box in the MP4.
- The `stss` box is now correctly omitted.


felicialim:
Tests for this are added at https://github.com/gpac/testsuite/pull/65

rbouqueau:
Thank you !

jeanlf:
Thanks ! Just curious here: is there a concept like 'roll' / SAP4 in iamf defined ?

felicialim:
> Thanks ! Just curious here: is there a concept like 'roll' / SAP4 in iamf defined ?

Yes there is (see https://aomediacodec.github.io/iamf/v1.1.0.html#audio_roll_distance).

In gpac, this should be set in https://github.com/gpac/gpac/blob/master/src/filters/reframe_av1.c#L1062.

PR #3681, opened 2026-07-04
Harden MP4Box dump filename construction

SABITHSAHEB:
While reviewing the MP4Box dump routines in applications/mp4box/filedump.c, I noticed several places where output filenames were assembled with unbounded formatting calls. Those paths were building scene-stat, cue, AV1/NAL, and timed-text file names directly from the user-provided base name.

That pattern could grow past the fixed stack buffers used in those functions and create a memory-safety risk. I patched the file to use bounded string construction instead, following GPAC’s existing safe string helper style and [snprintf](vscode-file://vscode-app/c:/Users/Kartik%20K/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) where formatting was needed.

The change keeps the same behavior for normal filenames, but removes the overflow-prone path when long names are used. It is a small hardening fix with low maintenance cost and no functional redesign.

### DIFF
--- a/applications/mp4box/filedump.c
+++ b/applications/mp4box/filedump.c
@@ -375,6 +375,7 @@ void dump_isom_scene_stats(char *file, char *inName, Bool is_final_name, u32 sta
 	dump = NULL;
 	sm = NULL;
 	sample_list = NULL;
+	e = GF_OK;
 
 	close = 0;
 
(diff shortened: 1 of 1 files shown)


applications/mp4box/filedump.c
```diff
@@ -400,12 +401,12 @@ void dump_isom_scene_stats(char *file, char *inName, Bool is_final_name, u32 sta
 	if (e<0) goto exit;
 
 	if (inName) {
-		gf_strcpy(szBuf, inName);
```
aureliendavid, line 403: If you look at https://github.com/gpac/gpac/blob/master/include/gpac/setup.h#L870 you can see that gf_strcpy and gf_strcat are macros that call the "l" version with sizeof() for local buffers, so the change is not need here 
other changes are ok, thanks

aureliendavid (commented):
Thanks for the contribution. 

Outside of the comment below (or above idk) the changes seem good to me.

applications/mp4box/filedump.c
SABITHSAHEB: Right, I'd missed that the macros already pass sizeof() for array buffers. Reverted those two lines back to gf_strcpy/gf_strcat, kept the rest.

PR #3682, opened 2026-07-05
isom: Fix integer truncation and missing allocation checks in box rea…

SABITHSAHEB:
While reviewing the box parser implementation, I noticed that several box readers (diST, dimC, and boxstring) cast the 64-bit s->size payload length to u32 during malloc. A large size wraps around, allocating a small buffer but writing the full amount, which leads to a heap buffer overflow. 

Additionally, mvcg_box_read was missing a NULL-pointer check on ptr->entries before memset.
I've added bounds checks (s->size <= 0xFFFFFFFE) to prevent the integer wrap-around and added validation checks for the allocations.

### DIFF
--- a/src/isomedia/box_code_3gpp.c
+++ b/src/isomedia/box_code_3gpp.c
@@ -1060,6 +1060,7 @@ GF_Err dimC_box_read(GF_Box *s, GF_BitStream *bs)
 	p->streamType = gf_bs_read_int(bs, 1);
 	p->containsRedundant = gf_bs_read_int(bs, 2);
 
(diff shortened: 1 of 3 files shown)


aureliendavid (changes requested):
Thanks for the contribution, this looks good to me. 

However instead of using plain `0xFFFFFFFF` it would be better to use `GF_UINT_MAX` for consistency with the rest of the code (and thus `GF_UINT_MAX-1` for `0xFFFFFFFE`, etc.)

SABITHSAHEB:
Makes sense, switched the three checks to GF_UINT_MAX-1.

PR #3684, opened 2026-07-06
webvtt: fix memory leaks and null pointer dereference on parse failure

SABITHSAHEB:
While going through the WebVTT parser functions, I noticed that gf_webvtt_parse_cues_from_data and gf_webvtt_dump_iso_sample leak the allocated bitstream, cues, and boxes if box parsing fails. Additionally, an allocation failure in gf_webvtt_cue_new causes a null pointer dereference when setting timestamps.

This patch introduces proper error checks and a structured goto cleanup target to safely free all allocated resources on error paths without breaking the public API.

### DIFF
--- a/src/media_tools/webvtt.c
+++ b/src/media_tools/webvtt.c
@@ -1250,19 +1250,25 @@ GF_List *gf_webvtt_parse_cues_from_data(const u8 *data, u32 dataLength, u64 star
 	GF_VTTCueBox *cuebox;
 	GF_BitStream *bs;
 	char *pre_text;
+	GF_Box *box = NULL;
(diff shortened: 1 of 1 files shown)


aureliendavid:
this one looks good, thanks

if you do the requested changes to the other two I'll merge the 3 PRs in one go

thanks for contributing !

SABITHSAHEB:
Both of the other PRs are updated now: #3682 uses GF_UINT_MAX for the size checks and #3681 keeps the gf_strcpy/gf_strcat macros as you pointed out. All three should be good to go.

aureliendavid:
thanks, all good

PR #3734, opened 2026-07-17
Use sysctl(KERN_PROC_PATHNAME) on FreeBSD instead of /proc/self/exe

ocochard:
/proc/self/exe requires procfs, which is not mounted by default on FreeBSD. The previous code stored readlink's ssize_t result into a u32 'size' variable, so a -1 error return became 0xFFFFFFFF, the size>0 check passed, and file_path[size] = 0 wrote out of bounds and crashed the process.

Use the native FreeBSD KERN_PROC_PATHNAME sysctl for the executable path lookup; it works without procfs. Fix the Linux path to store the readlink result in ssize_t and log a clear error mentioning procfs when the read fails.

### DIFF
--- a/src/utils/os_config_init.c
+++ b/src/utils/os_config_init.c
@@ -334,6 +334,12 @@ static Bool get_default_install_path(char file_path[GF_MAX_PATH], u32 path_type)
 /*Linux, OSX, iOS*/
 #else
 
+#include <errno.h>
+
+#if defined(__FreeBSD__)
+#include <sys/sysctl.h>
+#endif
+
 //dlinfo
 #if defined(__DARWIN__) || defined(__APPLE__) || defined(__FreeBSD__)
 #include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


rbouqueau:
Thanks for your contribution.

@aureliendavid @jeanlf Didn't we say last time that we wanted to add some BSDs to our build system?

ocochard:
Do you need an example of some `‎.github/workflows/ci-freebsd.yml` ?

rbouqueau:
We'll tell you, thanks for proposing. I see https://github.com/gpac/gpac/pull/3424#issuecomment-3774023585 from your previous PR. Let's wait for what the others think (probably not today).

aureliendavid:
thank you for the freebsd support! 

we have a freebsd build workflow running on master and on PRs that are built by buildbot (may need manual approval to run)

it ran successfully for this PR here: https://github.com/gpac/gpac/actions/runs/29733345553/job/88322905261

merged

PR #3389, opened 2025-11-10
mdcv av1 to mpeg: fix color primary order

rbouqueau:
see #3331

Issue #3331, opened 2025-09-08
Issue with AV1 to MPEG mdcv data transformation


MingxinChen:
Hi GPAC team,

I noticed an issue with the transformation function of mdcv data from AV1 to MPEG.

According to **ISO/IEC 14496-12:2022 (section 12.1.7)**, the data in mdcv is functionally equivalent to, and shall be as described in, the Mastering Display Colour Volume SEI message in H.265. Referring to **ITU-T H.265 v7 (11/2019) (section D.3.28)**, the values are defined as follows for display_primaries_x[c] and display_primaries_y[c]:
	•	c = 0 → Green primary
	•	c = 1 → Blue primary
	•	c = 2 → Red primary

However, in the **AV1 Bitstream & Decoding Process Specification 2019-01-08 (section 6.7.4)**, primary_chromaticity_x[i] and primary_chromaticity_y[i] are defined with indices:
	•	i = 0 → Red
	•	i = 1 → Green
	•	i = 2 → Blue

This means that the function should rearrange the order from RGB to GBR to align with the specification.
`void gf_av1_format_mdcv_to_mpeg(u8 mdcv_in[24], u8 mdcv_out[24])`

rbouqueau:
I checked the specs.

HEVC 2023/09 states that "For describing mastering displays that use red, green, and blue colour primaries, it is suggested that index value c equal to 0 should correspond to the green primary, c equal to 1 should correspond to the blue primary, and c equal to 2 should correspond to the red colour primary (see also Annex E and Table E.3)". Note the "should" not "shall".

All the listed `colour_primaries` syntax element are all indeed GBR, but there are some reserved we don't know about.

@jeanlf It seems to me that:
- We should indeed change the order in `gf_av1_format_mdcv_to_mpeg()`, I'm doing a PR.
- `mdcv` from the H265 SEI may be complex.

rbouqueau:
SOrry for the delay, I wanted to get some feedback. Colleague says it's ok but warns about double checking some tests.

jeanlf:
any news on this ?

rbouqueau:
I still planned to do it when I have some time.

rbouqueau:
It's been merged. Closing. Thanks again @MingxinChen !

PR #3757, opened 2026-07-22
Fix AV1 MDCV component ordering

mlrsmith:
in `gf_av1_format_mdcv_to_mpeg`, used when parsing an AV1 MDCV metadata OBU to convert to the format used in an mp4 `mdcv` box (and thus used when using MP4Box to package an av1 obu bitstream to mp4), there is a comment describing, correctly, the reordering of the colour components.

However, the code does not match that ordering. This PR changes it to match the comment.


rbouqueau:
Hi, thanks for the report. The previous effort was #3331 reported by Dolby and with attempt to be solved by #3389. In addition a clarification contribution was sent to MPEG ([issue](https://git.mpeg.expert/MPEG/Systems/FileFormat/isobmff/-/issues/420)) but I see this was closed next week - I will ask why.

@mlrsmith Would you be able to share a sample + command-line for your issue? Do you confirm your GPAC is recent enough to contain the aforementioned fix?

podborski:
IIRC
Mdcv in AV1 is not compliant to SMPTE spec as it also uses different precision,
while ISOBMFF is according to SMPTE spec precision. I recall this is some kind of VP8 artifact.

mlrsmith:
We found this bug when using gpac v26.02.0, which contains both  https://github.com/gpac/gpac/pull/3389 and the follow-up fix to that https://github.com/gpac/gpac/commit/7b2fc0b347bff982e816853645a927997f383acb

In this code, the shuffle array `{4,5,0,1,2,3,6,7}` is the ordering of the _source_ components used to produce the output. Since the source here is in RGB order, this means the output is BRG, not (as intended) GBR. This PR changes the shuffle indexes to produce GBR.

The specific command line we're using is:
```
MP4Box -add 'input.obu:fps=30000/1001:dvp=10.hdr10:negctts' -new output.mp4
```

Though the additional input options here are not directly relevant to this bug.

I don't have any sample files that I can freely redistribute, though I can create a fake one easily enough if that's important ("fake" in that it's just encoded with arbitrary MDCV values that aren't particularly connected to the source material).

mlrsmith:
[bear-av1-hdr.obu.gz](https://github.com/user-attachments/files/30283292/bear-av1-hdr.obu.gz)

Attached file is an OBU bitstream (gzipped because github wouldn't let me attach it otherwise). The source media for this was a test file from the shaka-packager git repository; it's been re-encoded with the HDR metadata attached (the underlying media is not _actually_ HDR though).

rbouqueau:
@mlrsmith Can you merge against an updated master branch here so that the tests run properly?

mlrsmith:
> @mlrsmith Can you merge against an updated master branch here so that the tests run properly?

Done, though it looks like someone else will need to allow the tests to run.

rbouqueau:
@mlrsmith LGTM, thanks. I can update the hashes and merge tomorrow if there is no objection.

mlrsmith:
Thanks! That would be great.

PR #3827, opened 2026-07-30
iamf: queue frames before pre-skip is finalized

felicialim:
Avoid dropping the initial frames parsed before the PID is created. Dropping them causes edit lists (elst) in the output MP4 to discard active audio samples instead of the intended delay padding.

### DIFF
--- a/src/filters/reframe_av1.c
+++ b/src/filters/reframe_av1.c
@@ -36,6 +36,14 @@ typedef struct
 	Double duration;
 } AV1Idx;
 
+typedef struct {
+	u8 *data;
+	u32 size;
+	u64 cts;
+	s32 roll_distance;
+	u32 trim_at_end;
+} GF_IamfQueuedFrame;
+
 typedef enum {
 	//input format not probed yet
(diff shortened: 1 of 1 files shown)


felicialim:
The associated test is at https://github.com/gpac/testsuite/pull/70

rbouqueau:
@felicialim I don't see the failing `mp4box-mux` test being updated at https://github.com/gpac/testsuite/pull/70/, may I ask you to check?

rbouqueau:
@felicialim Otherwise it is ok, thanks again. While reviewing, we were wondering about splitting from the AV1 and IAMF codes, what do you think? We have a script for [splitting while keeping history](https://github.com/gpac/gpac/tree/master/.github/scripts).

felicialim:
https://github.com/gpac/testsuite/pull/70 updated, sorry about that! 

Splitting the AV1 and IAMF code makes sense to me too.

rbouqueau:
Thank you. Do you wish to split the code? Otherwise I'll put it on my todo.

PR #3853, opened 2026-08-10
ffmpeg9

rbouqueau:
builds since ffmpeg6 where other non-ifdef-ed errors happen anyway (v7 and v8 ok)

### DIFF
--- a/configure
+++ b/configure
@@ -1717,7 +1717,7 @@ config_package faad "faad2" "" "-lfaad -lm" "" '#include <faad.h>
 int main( void ) { NeAACDecOpen(); return 0; }'
 
 
-ffmpeg_lflags="-lz -lavcodec -lavformat -lavutil -lswscale -lpostproc"
+ffmpeg_lflags="-lz -lavcodec -lavformat -lavutil -lswscale"
 ffmpeg_pck_names="libavcodec libavformat libswscale libavutil"
 
 if test "$is_em" = "yes" ; then
--- a/src/filters/ff_common.h
+++ b/src/filters/ff_common.h
@@ -68,6 +68,10 @@
 #define FFMPEG_OLD_CHLAYOUT
 #endif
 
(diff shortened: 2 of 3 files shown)


aureliendavid:
> builds since ffmpeg6 where other non-ifdef-ed errors happen anyway (v7 and v8 ok)

I'm not sure I understand what this means? 

this probably needs at least some ifdefs or refactor to be compatible with ffmpeg versions that don't have `avcodec_get_supported_config()` (which was added in 7.1 I think)

rbouqueau:
Actually I wrote a small script for retro compat which passed with FFmpeg version down to 6.0. But now reproducing by hand in the light of the buildbot result, I've realized that GPAC failed at detecting some older versions of FFmpeg.

aureliendavid:
I think we based most compatibility guards on the relevant LIBAVXXX_VERSION_MAJOR macros

Ideally we'd like to be retro-compatible down to ffmpeg 4 I think, the only hard stops I see in the code are: 

https://github.com/gpac/gpac/blob/d34570e3d994ede1f7fa84075fea793f8939308e/src/filters/ff_mx.c#L34-L36

(57 is ffmpeg 3)

and 

https://github.com/gpac/gpac/blob/d34570e3d994ede1f7fa84075fea793f8939308e/src/filters/ff_avf.c#L36-L38

(7 is ffmpeg 4)

rbouqueau:
I checked and n4.2.9 is the latest FFmpeg tag I can build. My compiler doesn't accept the code before that (and I don't plan to get an older version to make further tests). Weirdly GPAC detects the build as 'custom' ;) - won't investogate that either.

aureliendavid:
if we can get support for v4.2.9 to v9 with just macro checks that'd be well enough

aureliendavid:
looks good now, will merge with the configure PR and other stuff before eod, ok with you @rbouqueau ?

rbouqueau:
Just reconnecting now. All good, thank you!

PR #3854, opened 2026-08-10
configure: fail when --use-XXX is used but XXX is not found

rbouqueau:
see #3853 CC @aureliendavid

### DIFF
--- a/configure
+++ b/configure
@@ -427,7 +427,7 @@ GPAC core configuration:
 
 
 Extra libraries configuration:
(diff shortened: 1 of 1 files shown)


rbouqueau:
/buildbot-rebuild

I put the wrong branch name, my bad. Multi-tasking too much :(

aureliendavid:
(the label is enough to trigger the build, no need to force a rebuild right away)

the PR seems reasonable to me

Issue #2811, opened 2024-04-26
Apple MV-HEVC ISOBMFF extensions


rbouqueau:
https://developer.apple.com/av-foundation/Stereo-Video-ISOBMFF-Extensions.pdf

rbouqueau:
Not sure it is 100% related but I found a back reference here: https://blog.mikeswanson.com/mv-hevc-with-x265-and-nvidia/.

PR #3857, opened 2026-08-17
isom: fix oinf/linf sample groups dropped when fragmenting

fkadiri-ybvr:
`gf_isom_add_sample_group_info_internal()` routes every new sample group description to the current traf as soon as fragmented writing is active (`FragmentsFlags & GF_ISOM_FRAG_WRITE_READY`), with no exception for `oinf` (Operating Points Information, used by MV-HEVC/L-HEVC) or `linf` (Layer Information). Both describe the track as a whole, the same as VPS/SPS/PPS, not a per-sample property like rap/roll/sync. Routing them to traf means they get duplicated into every fragment and never written to the init segment's moov/stbl. Any player that reads only the init segment, which is how HLS/CMAF clients are meant to work via EXT-X-MAP, never sees the operating-points declaration and can't recognize the track as valid layered HEVC, even though the underlying bitstream is fine.

Separately, `mp4_mux_set_hevc_groups()`'s scan for an external HEVC base track doesn't exclude the current track. For a single combined base+enhancement track (sample entry hvc1/hev1, both layers muxed together, which is what a number of MV-HEVC encoders produce today), the scan matches the track against itself, sets a self-referential OREF/BASE track reference, and removes the oinf group that had just been attached two lines above. linf isn't touched by that removal, which is why only oinf goes missing in that case.

Fix: oinf and linf are exempted from the fragmented-write routing in `gf_isom_add_sample_group_info_internal()` and always attach to the sample table. The base-track scan in `mp4_mux_set_hevc_groups()` skips the current track.

Verified against a real MV-HEVC asset packaged for HLS (onDemand profile): isolating just the EXT-X-MAP init segment bytes now shows the full Operating Points Information box (scalability_mask Multiview, dependency_layers 2), matching a non-fragmented reference file produced from the same source. `-unit-tests` passes. Also spot-checked a plain non-layered file through the same fragmented pipeline for regressions; no structural issues turned up, though this hasn't been run against the full testsuite corpus.

### DIFF
--- a/src/filters/mux_isom.c
+++ b/src/filters/mux_isom.c
@@ -8355,7 +8355,10 @@ static void mp4_mux_set_hevc_groups(GF_MP4MuxCtx *ctx, TrackWriter *tkw)
 	}
 	//set linf
 	for (i=0; i<gf_isom_get_track_count(ctx->file); i++) {
-		u32 subtype = gf_isom_get_media_subtype(ctx->file, i+1, 1);
+		u32 subtype;
+		//a track is never its own external base layer
+		if (i+1 == tkw->track_num) continue;
+		subtype = gf_isom_get_media_subtype(ctx->file, i+1, 1);
 		if ( gf_isom_is_media_encrypted(ctx->file, i+1, 1))
 			gf_isom_get_original_format_type(ctx->file, i+1, i+1, &subtype);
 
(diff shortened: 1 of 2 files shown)


fkadiri-ybvr:
@rbouqueau would appreciate your take on this, given your earlier comment on #2811.

rbouqueau:
Thanks. I'm interested in case people need it (that's your case, correct?) but I think @jeanlf is our expert here.

fkadiri-ybvr:
Correct, this is our case. We hit it packaging MV-HEVC for HLS delivery to Vision Pro and iPhone: `oinf` was landing only in fragments, never in the init segment, so any player reading just the init segment (the normal HLS/CMAF flow via EXT-X-MAP) couldn't recognize the track as layered HEVC. Fixed and verified against a real fragmented MV-HEVC asset, happy to answer anything or adjust the approach if @jeanlf sees a cleaner way to do it.

PR #3888, opened 2026-09-03
Fix the assignment of n_signals in AC-4 parser

MingxinChen:
In the current implementation, `n_signals` is incorrectly calculated by hardcoding `b_lfe = 1` as a default value. According to _ETSI TS 103 190-2 V1.3.1 (2025-07)_ Table 60, the correct mapping of n_signals should be `[0,1,2,3,5] + b_lfe`.


rbouqueau:
All good, thank you!

PR #3890, opened 2026-09-06
SCTE-35: handle unspecified splice command length sentinel

JungleGenius:
# Fix SCTE-35 unspecified splice_command_length (0xFFF)

## Problem

`splice_command_length` is a 12-bit field. A value of `0xFFF` indicates that the command length is unspecified rather than that 4095 command bytes follow. The current decoder treats `0xFFF` literally, rejects otherwise valid broadcast SCTE-35 sections as truncated, and for `splice_null()` attempts to skip 4095 bytes.

The issue was reproduced with a real MPEG-TS SCTE-35 `splice_insert()` section that uses `splice_command_length=0xFFF`.

## Fix

Treat `0xFFF` as the unspecified-length sentinel:

- skip the literal available-bytes validation for that sentinel;
- do not skip 4095 bytes for `splice_null()` when the length is unspecified;
- leave all explicitly sized command behavior unchanged.

## Unit test

Adds `scte35dec_unspecified_splice_command_length` using the real 50-byte broadcast section. The test verifies that parsing succeeds and yields:

- splice event ID: `662`
- PTS: `8076794552`
- break duration: `21780000` ticks (`242` seconds at 90 kHz)
- IDR requirement: true

## Rationale

This is a parser correctness fix. It does not alter SCTE-35 timing, event conversion, HLS/DASH output.

## Review notes

The patch is deliberately small so it can be reviewed independently from the other SCTE/HLS findings.

### DIFF
--- a/src/filters/dec_scte35.c
+++ b/src/filters/dec_scte35.c
@@ -478,7 +478,9 @@ Bool scte35dec_get_timing(const u8 *data, u32 size, u64 *pts, u64 *dur, u32 *spl
 	/*u32 tier = */gf_bs_read_int(bs, 12);
 
 	u32 splice_command_length = gf_bs_read_int(bs, 12);
-	if (splice_command_length > gf_bs_available(bs)) {
+	/* 0xFFF indicates that splice_command_length is unspecified. */
+	Bool splice_command_length_unspecified = (splice_command_length == 0xFFF);
+	if (!splice_command_length_unspecified && (splice_command_length > gf_bs_available(bs))) {
 		GF_LOG(GF_LOG_ERROR, GF_LOG_CODEC, ("[Scte35Dec] Bitstream too short (" LLU " bytes) while parsing splice command (%u bytes)\n",
 			gf_bs_available(bs), splice_command_length));
 		goto exit;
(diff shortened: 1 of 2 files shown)


rbouqueau:
Thx. Looks like a net improvement. Can you add a comment on [src/filters/unittests/ut_dec_scte35.c](https://github.com/gpac/gpac/pull/3890/changes#diff-8b0123945a8329193f85f3aa4a129efc6efad7007cdb375a6741f808f3540751) is constructed? It is not obvious this is a `splice_null()` followed by anything just seeing the hex dump.

JungleGenius:
Hi @rbouqueau I added it to the PR source. I expanded the comment above the captured test vector to decode the relevant SCTE-35 fields so the purpose of the hex payload is clear without having to reverse-engineer it.

One clarification from decoding the sample: this particular captured section is a `splice_insert` (`splice_command_type=0x05`) with `splice_command_length=0xFFF` (unspecified), rather than a `splice_null`. The comment now documents the command type, event ID, splice time, and break duration as well as the reason for the regression test.

Let me know if you need anything else on this PR.

rbouqueau:
LGTM. Launched the buildbot and will merge.

rbouqueau:
Merging, thanks again !

PR #3891, opened 2026-09-06
HLS: normalize SCTE-35 time against presentation time offset

JungleGenius:
# Normalize SCTE-35 time for HLS when presentationTimeOffset is non-zero

## Problem

GPAC can parse SCTE-35 from MPEG-TS and expose it as an MPD event stream, but HLS SCTE manifest generation compares two different clock domains when the source has a non-zero MPEG-TS timeline:

- SCTE `splice_time()` remains on the source media clock;
- HLS segment contexts are presentation-relative after DASH `presentationTimeOffset` (PTO) is applied.

As a result, valid events are not associated with any HLS segment and no `EXT-X-DATERANGE` / `EXT-X-CUE-OUT` marker is emitted.

Observed real-source values:

- SCTE event PTS: `8076794552`
- PTO: `8070463950`
- normalized event time: `6330602` ticks (`70.340022 s`)

DASH EventStream output already demonstrated that the cue itself was parsed correctly; the failure was limited to HLS segment/event matching.

## Fix

Resolve the representation's inherited segment timescale and PTO, then:

- subtract PTO from source-clock SCTE event time when appropriate;
- rescale event time and duration to the representation timescale for segment matching;
- use the normalized presentation time for HLS `START-DATE` calculation.

## Unit test

Adds `mpd_hls_scte35_nonzero_presentation_time_offset` using the real timing values. It verifies that the event is matched to the expected presentation-relative segment and that the HLS writer emits:

- `#EXT-X-DATERANGE:ID="662-0000"`
- `#EXT-X-CUE-OUT:242`

## Validation

In the integration test, the same real SCTE source generated DATERANGE, CUE-OUT, and CUE-IN after the fix. The generated CMAF media fragments were byte-for-byte identical before and after the change; only manifest SCTE timing/signaling changed.

## Rationale

HLS event matching should occur in one clock domain. Normalizing source-clock SCTE timing by the inherited PTO is consistent with the presentation-relative segment state already used by the HLS writer.

### DIFF
--- a/src/media_tools/mpd.c
+++ b/src/media_tools/mpd.c
@@ -4325,17 +4325,38 @@ static void hls_insert_crypt_info(FILE *out, GF_MPD_Representation *rep, GF_DASH
 	}
 }
 
-static void hls_insert_scte35_info(FILE *out, u64 ast, const GF_MPD_Period *period, GF_DASH_SegmentContext *sctx)
+static void hls_insert_scte35_info(FILE *out, u64 ast, GF_MPD_Period *period, GF_MPD_AdaptationSet *adaptation_set, GF_MPD_Representation *representation, GF_DASH_SegmentContext *sctx)
 {
+	u64 segment_duration = 0;
+	u64 presentation_time_offset = 0;
+	u32 segment_timescale = 1;
 	GF_MPD_EventStream *es = NULL;
+
+	/* Segment contexts are presentation-relative, while SCTE-35 splice_time is
+	 * carried on the source timeline. Resolve the inherited PTO once so both
+	 * values are compared in the same clock domain. */
+	gf_mpd_resolve_segment_duration(representation, adaptation_set, period,
+		&segment_duration, &segment_timescale, &presentation_time_offset, NULL);
+	if (!segment_timescale) segment_timescale = 1;
 	u32 i = 0;
 	while ( (es = gf_list_enum(period->event_streams, &i)) ) {
 		GF_MPD_EventStreamEntry *ese = NULL;
 		u32 j = 0;
 		while ( (ese = gf_list_enum(es->entries, &j)) ) {
-			if (ese->state == 0 && sctx->time <= ese->presentation_time && ese->presentation_time < sctx->time+sctx->dur) {
+			u64 event_time = ese->presentation_time;
+			u64 event_time_in_representation_timescale;
+			u64 event_duration_in_representation_timescale;
+			u64 presentation_time_offset_in_event_timescale = gf_timestamp_rescale(presentation_time_offset, segment_timescale, es->timescale);
+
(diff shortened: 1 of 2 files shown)


rbouqueau:
LGTM. I've asked the buildbot to run on this first. We definitely lack SCTE35 real life examples. Anything that we could extract for our testsuite would be great!

JungleGenius:
Thanks for the quick response.

Just an FYI - We were working on a live SCTE-35 pass thru and found several minor edge cases issues along the way that we thought may have been overlooked.

We broke them all into separate Pull Requests so they would be easier to evaluate individually.

However, I have three other remaining Pull Requests from this same sprint, but one of them is dependent on these getting merged first otherwise it would conflict so I haven't made the other three Pull Requests yet.

I can get you a link the SCTE-35 source we are testing with later today. It's a well known example used by many other video projects on Github so it's almost unofficially a canonical example.

-Daniel

Daniel Thompson
  Jungle Genius LLC
  PO Box 2591
  Goldenrod FL 32733
  (407) 592-4365

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

JungleGenius:
Quick update: I’ve reproduced the two `scte35-dasher` HashFails and confirmed that they are the expected playlist changes from applying the 80 ms presentation-time offset. The generated media/init output remains unchanged.

I still need to generate the corresponding `gpac/testsuite` reference hashes from a clean build of this PR branch itself (`fix/02-hls-scte35-presentation-time-offset`), rather than from our internal production build.

The EC2 I use for GPAC builds is currently offline, so I’ll bring that back up tomorrow and prepare the same-named testsuite branch with the updated hashes as soon as possible.

rbouqueau:
@JungleGenius Thanks for the testsuite. I also expected this diff. If the output looks good to you, you can do `cd testsuite && ./make_tests.sh -clean scripts/scte35.sh && ./make_tests.sh -hash scripts/scte35.sh` to regenerate the appropriate hash.

Thanks for the PRs. Waiting for the sample. If your contributions become significant, we'll need you to sign a CLA, let's do this offline.

rbouqueau:
Can you please make a PR for the tests? Should be quite straight-forward if you follow my instructions. We should be good after that, thanks again.

JungleGenius:
The companion testsuite PR is now open:

[https://github.com/gpac/testsuite/pull/77](https://github.com/gpac/testsuite/pull/77)

It updates only the two expected SCTE-35 HLS reference hashes for:

* `scte35-dasher-hls-simple`
* `scte35-dasher-hls-multi`

The hashes were regenerated using GPAC built directly from `fix/02-hls-scte35-presentation-time-offset`, not our cumulative production branch.

This should provide the matching testsuite branch needed to rerun Buildbot for this PR.

rbouqueau:
Thank you. Merging.

PR #3892, opened 2026-09-06
SCTE-35: schedule passthrough cues at first suitable RAP

JungleGenius:
# Make scte35dec passthrough schedule cue points on usable random-access packets

## Problem

`scte35dec:mode=passthrough` is intended to pass media while adding `CueStart` at SCTE splice points. Two implementation details prevent that behavior from working reliably:

1. passthrough returns from PID configuration before determining the SCTE input data format;
2. SCTE events are only fed into the internal event scheduler in EVTE mode, so passthrough can wait for an event that was never scheduled.

A second timing issue is exposed by real broadcast material: an SCTE splice time can precede the next usable video random-access packet by a small amount. Requiring packet CTS to equal splice CTS exactly can therefore miss the intended splice point.

Observed real-source relationship:

- SCTE splice PTS: `8076794552`
- next video RAP: `8076796350`
- delta: `1798` ticks (~19.98 ms at 90 kHz)

## Fix

- determine input data mode before the passthrough early return;
- feed SCTE data into the existing event scheduler in passthrough mode as well as EVTE mode;
- keep a cue pending on predictive packets;
- consume/mark the cue on the first packet with a SAP at or after the scheduled splice CTS.

## Unit test

Extends the existing splice-point test to verify that:

- a non-SAP packet at the splice timestamp does not consume the event;
- the first SAP packet after the event does consume it and becomes the cue point.

## Rationale

For stream-copy/pass-through operation, GPAC cannot create a new IDR at an arbitrary SCTE timestamp. The usable segment boundary is therefore the first existing random-access point at or after the cue. This preserves decodeability while respecting the signaled break as closely as the source encoding permits.

## Scope

This patch does not itself change dasher segmentation policy. It only makes the documented passthrough cue annotation path functional and correctly associates a scheduled cue with a usable packet.

### DIFF
--- a/src/filters/dec_scte35.c
+++ b/src/filters/dec_scte35.c
@@ -169,7 +169,6 @@ static GF_Err scte35dec_configure_pid(GF_Filter *filter, GF_FilterPid *pid, Bool
 
 	//copy properties at init or reconfig
 	gf_filter_pid_copy_properties(ctx->opid, pid);
-	if (ctx->mode == PASSTHRU) return GF_OK;
 
 	const GF_PropertyValue *p = gf_filter_pid_get_property(ctx->ipid, GF_PROP_PID_CODECID);
 	if (p) {
@@ -178,6 +177,8 @@ static GF_Err scte35dec_configure_pid(GF_Filter *filter, GF_FilterPid *pid, Bool
 		else if (p->value.uint == GF_CODECID_EVTE)
 			ctx->data_mode = BOX;
 	}
+
+	if (ctx->mode == PASSTHRU) return GF_OK;
 
 	gf_filter_pid_set_property(ctx->opid, GF_PROP_PID_STREAM_TYPE, &PROP_UINT(GF_STREAM_METADATA) );
 	if (ctx->mode == M2TS_SEC) {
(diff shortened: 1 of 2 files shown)


rbouqueau:
Thanks. Very useful to have more real life tests on this.

The patch is probably not ok because I could revert almost all proposed code and get the UT working + we don't want `PASSTHRU` to actually process and dispatch data.

If the problem is only some CTS accuracy issues (which the UT proposal doesn't cover) then I think it'd be ok to just modify this condition: `Bool is_splice = (evt->dts + evt->emib->presentation_time_delta == cts);` with a threshold.

19.98 ms seems a lot to be honest. Maybe it is some interlaced content?

I don't understand what the other problem is (i.e. why we'd need to know the input format for passthru). At the same time, I can recognized the current approach is simplistic because it just put a CUE when it sees a SCTE35 payload even if it's a `splice_null`. Any idea or remark?

Is there any sample you could share with us in our [this dropbox](https://www.mediafire.com/filedrop/filedrop_hosted.php?drop=eec9e058a9486fe4e99c33021481d9e1826ca9dbc242a6cfaab0fe95da5e5d95). Seeing the actual issue may help.

JungleGenius:
Hi @rbouqueau , I reworked the implementation based on your feedback.

PASSTHRU no longer creates/processes or dispatches Event Message boxes, and the additional input-format handling has been removed. It remains a pass-through path and only examines the existing SCTE-35 packet property far enough to determine whether there is an actionable splice.

`splice_null` no longer creates a pending cue. For an actionable splice, the target is retained and an appropriate SAP/RAP can satisfy it when it falls within one packet duration of the signaled timestamp.

I also checked your interlace hypothesis against the real source and you were right. The sample is top-field-first interlaced H.264 at 25 frames/s / 50 fields/s. The observed offset is 1798 ticks on the 90 kHz clock, or 19.978 ms, essentially one 20 ms field.

The revised regression test uses the actual 1798-tick case, includes a `splice_null` case, and verifies that a RAP outside the one-packet-duration tolerance is not accepted.

I also reran our cumulative/real-source validation. The revised implementation still selects the same video RAPs we relied on previously, preserves all 183/183 SCTE-35 packets byte-for-byte through our metadata filter, and leaves our independent timed-ID3/emsg path unchanged.

The real-world source is the longstanding FFmpeg SCTE-35 sample:

https://samples.ffmpeg.org/ffmpeg-bugs/trac/ticket3356/mpegwithscte35.ts

The same sample has also been referenced or used for SCTE-35 testing by GStreamer, Elecard, and threefive.

rbouqueau:
Ok saw the sample here. Thx.

Will review the rest later too.

rbouqueau:
LGTM, thanks for the update. Launched the buildbot to be able to merge afterward.

PR #3906, opened 2026-09-14
Associate HLS SCTE-35 cues with segment start boundaries

JungleGenius:
## Problem

The HLS SCTE-35 writer currently associates an event with the segment whose time range contains the event. That can place `EXT-X-CUE-OUT` before a segment that began before the actual splice point, and similarly place `EXT-X-CUE-IN` before a segment that still contains media preceding the return point.

For SSAI, cue tags are segment-boundary signals. Starting replacement at a segment that began before the SCTE splice time replaces valid program media too early.

A simple example is:

- segment: `10.0 -> 14.0 s`
- SCTE break start: `10.3 s`

The current range test emits CUE-OUT before the 10.0 s segment. The first existing HLS boundary at or after the event is 14.0 s, so absent cue-driven media segmentation the marker belongs before that segment instead.

The same issue applies to break end. If the return point is 15.3 s inside the `14.0 -> 18.0 s` segment, CUE-IN belongs before the segment beginning at 18.0 s rather than before the 14.0 s segment.

## Fix

Treat HLS SCTE cue tags as boundary-oriented signaling:

- emit `DATERANGE` / `CUE-OUT` on the first segment whose start is at or after the SCTE event time while the break is active;
- emit `CUE-IN` on the first segment whose start is at or after the event end;
- keep `DATERANGE START-DATE` at the exact SCTE event time.

This patch does **not** create or move media segment boundaries. Cue-driven segmentation is intentionally handled in a separate change.

## Unit test

Adds `mpd_hls_scte35_cues_follow_segment_boundaries`.

The test verifies that:

- no CUE-OUT is emitted on a segment that started before a splice occurring inside it;
- CUE-OUT is emitted on the next segment boundary;
- no CUE-IN is emitted on a segment that contains the return point;
- CUE-IN is emitted on the first segment boundary after the return point.

## Validation

This standalone patch applies cleanly to GPAC base commit `239c20cdfe9dad19580e87f98ca0fd8b55fe560b`.

In the cumulative SCTE fix tree, this behavior is combined with the PTO-normalization fix so source-clock SCTE timestamps are first converted to the presentation timeline and then associated with HLS boundaries.

## Rationale

This change separates two concerns:

1. the HLS writer should never advertise a splice before media that precedes the splice;
2. the segmenter may optionally create a closer boundary at a suitable RAP.

Keeping those responsibilities separate makes the manifest behavior correct even when cue-driven segmentation is not enabled.

### DIFF
--- a/src/media_tools/mpd.c
+++ b/src/media_tools/mpd.c
@@ -4354,7 +4354,13 @@ static void hls_insert_scte35_info(FILE *out, u64 ast, GF_MPD_Period *period, GF
 			event_time_in_representation_timescale = gf_timestamp_rescale(event_time, es->timescale, representation->timescale);
 			event_duration_in_representation_timescale = gf_timestamp_rescale(ese->duration, es->timescale, representation->timescale);
 
-			if (ese->state == 0 && sctx->time <= event_time_in_representation_timescale && event_time_in_representation_timescale < sctx->time+sctx->dur) {
+			/* HLS cue tags describe segment boundaries. If an SCTE event falls
+			 * inside a segment, associate CUE-OUT with the first segment that
+			 * starts at or after the splice point rather than with the segment
+			 * that started before it. */
+			if (ese->state == 0
+				&& event_time_in_representation_timescale <= sctx->time
+				&& sctx->time < event_time_in_representation_timescale + event_duration_in_representation_timescale) {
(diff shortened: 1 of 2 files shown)


rbouqueau:
LGTM also, thank you.

PR #3907, opened 2026-09-14
Add SCTE-35 RAP boundaries without replacing normal segment cadence

JungleGenius:
## Problem

For HLS SSAI, translating SCTE-35 to manifest tags is not sufficient when the splice falls inside an already-formed media segment. The media needs a usable segment boundary near the splice point.

GPAC already supports `DashCue=inband`, but that mode is cue-driven segmentation: normal duration-based segment boundaries are suppressed while the segmenter waits for cue points. With a six-second target and a first SCTE cue around 70 seconds, using cue-only segmentation can therefore create an initial segment roughly 70 seconds long.

That is not the desired SCTE behavior. SCTE should add a splice boundary to the normal HLS cadence, not replace the cadence.

A second practical detail is that the SCTE splice timestamp may be slightly before the next encoded random-access packet. In the real broadcast sample used to reproduce the issue:

- normalized SCTE OUT time: `70.340022 s`
- next usable video RAP: `70.360 s`
- delta: about `19.98 ms`

Stream-copy packaging cannot manufacture an IDR at 70.340022 s. The first decodable segment boundary is the RAP at 70.360 s.

## Fix

When dasher parses a new SCTE event that requires an IDR/RAP on a visual representation:

- retain normal duration-based segmentation;
- remember the break start and, when a finite break duration is present, the planned return time;
- keep each boundary pending through predictive packets;
- force an additional segment boundary on the first SAP at or after the SCTE boundary;
- re-anchor the normal segment cadence at that SAP so following segments retain normal target-duration behavior;
- leave explicit source-driven, in-band-cue-only, and cue-list segmentation modes unchanged.

Repeated copies of the same SCTE event are already de-duplicated by the existing event-ID handling and therefore do not repeatedly reset the pending boundary.

## Unit test

Adds `dasher_scte35_boundary_uses_first_sap_at_or_after_splice`.

The test verifies that:

- a predictive packet after the splice is not used as a boundary;
- a SAP before the splice is not used;
- a SAP exactly at the splice is accepted;
- the first SAP slightly after the splice is accepted;
- comparison remains correct when packet and SCTE clocks use different timescales.

## Integration validation

Using the real MPEG-TS SCTE-35 sample with a six-second HLS target:

- SCTE OUT: `70.340022 s`
- new segment start: `70.360 s`
- CUE-OUT: immediately before that segment
- planned SCTE return: `312.340022 s`
- return segment start: `312.360 s`
- CUE-IN: immediately before that segment

Normal segmentation remained active before, during, and after the break. Around the first cue, the video segment timeline was:

- `61.480 -> 67.840`
- `67.840 -> 70.360` (shortened to reach the splice RAP)
- `70.360 -> 76.720`
- `76.720 -> 83.080`

Around the return:

- `299.320 -> 305.680`
- `305.680 -> 312.360`
- `312.360 -> 318.720`

## Rationale

SCTE signaling describes an intended splice time, while independently decodable HLS segments must begin on a usable random-access point. Choosing the first SAP at or after the cue avoids replacing program material early and preserves decodeability without transcoding.

Re-anchoring the normal cadence at the splice prevents the forced boundary from creating an immediately short follow-on segment.

## Scope

This is deliberately not a general segmentation-policy rewrite. It only adds an SCTE-derived RAP boundary when GPAC is otherwise using ordinary duration-based segmentation.

### DIFF
--- a/src/filters/dasher.c
+++ b/src/filters/dasher.c
@@ -523,6 +523,13 @@ typedef struct _dash_stream
 	s32 cues_ts_offset;
 	Bool inband_cues;
 
+	/* SCTE-35 splice boundaries are expressed on the MPEG 90 kHz clock.
+	 * Keep the start and optional return boundary while normal duration-based
+	 * segmentation continues. */
+	u64 scte35_splice_start;
+	u64 scte35_splice_end;
+	u8 scte35_boundary_state; /* 0: none, 1: break start, 2: break end */
+
 	Bool clamp_done;
 	u32 dcd_not_ready;
 
@@ -9920,6 +9927,12 @@ static void dasher_set_pto(GF_DashStream *ds, u64 pto_adj)
 	}
 }
 
+GF_STATIC Bool dasher_scte35_boundary_due(u64 packet_cts, u32 packet_timescale, u32 sap_type, u64 boundary_time)
+{
+	if (!sap_type || !packet_timescale) return GF_FALSE;
+	return gf_timestamp_greater_or_equal(packet_cts, packet_timescale, boundary_time, 90000);
+}
+
 static GF_Err dasher_handle_scte35(GF_DasherCtx *ctx, GF_FilterPacket *pck, GF_DashStream *ds)
 {
 	GF_Err ret = GF_OK;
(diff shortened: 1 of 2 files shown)


rbouqueau:
LGTM, thanks. Waiting for the CLA signature to merge.

PR #3908, opened 2026-09-14
Preserve SCTE-35 markers across repeated HLS playlist serialization

JungleGenius:
## Problem

The HLS SCTE-35 writer uses `GF_MPD_EventStreamEntry::state` to track whether a cue-out has been emitted while walking the segments of one playlist. The state is stored on the MPD event entry itself and therefore survives after that playlist serialization finishes.

This becomes incorrect when the same period/event list is serialized again, as happens for dynamic HLS playlist rewrites and for multiple HLS variant playlists sharing the same event stream. A cue emitted during the first write can leave the event in the "already emitted" state, causing a later full playlist serialization to omit the same `EXT-X-DATERANGE` / `EXT-X-CUE-OUT` marker even though the applicable segment is still present.

The issue was observed in a live HLS stream: the SCTE decoder reported the real splice event, but a later dynamic playlist rewrite no longer contained the cue marker. Static one-shot serialization did not show the problem.

## Root cause

`GF_MPD_EventStreamEntry::state` is serialization scratch state, but it is retained across calls to `gf_mpd_write_m3u8_playlist()`.

The state is needed while walking the segments of one playlist so the writer can pair CUE-OUT and CUE-IN. It must not carry the result of a previous playlist write into the next complete serialization.

## Fix

Reset event-entry serialization state at the start of each HLS variant playlist write, before walking that playlist's retained segment list.

The change is deliberately limited to HLS playlist serialization. It does not alter SCTE parsing, event timing, media segmentation, DASH output, or media payloads.

## Unit test

Adds `mpd_hls_scte35_dynamic_rewrite_preserves_cues`.

The test creates one SCTE event whose break extends beyond the current segment and serializes the same dynamic HLS playlist twice:

1. the first write emits `#EXT-X-CUE-OUT:20` and leaves the event state active while that playlist is being walked;
2. the second full playlist write must emit the same CUE-OUT again because it is a new serialization of the retained media window.

With the test added but the fix removed, the second assertion fails because the cue is suppressed by the state left from the first write. With the fix present, both writes contain the cue.

## Validation

Against base commit `239c20cdfe9dad19580e87f98ca0fd8b55fe560b`:

## Rationale

A full HLS playlist rewrite must be deterministic from the current MPD/event and retained-segment state. The result should not depend on whether another variant or an earlier dynamic rewrite happened to serialize the same event first.

## Scope

This patch fixes only repeated HLS serialization state. Cue timing normalization, cue/segment association, and SCTE-driven segment boundaries are separate changes in the patch series and are intentionally not included here.

### DIFF
--- a/src/media_tools/unittests/ut_mpd.c
+++ b/src/media_tools/unittests/ut_mpd.c
@@ -5,6 +5,86 @@
 #include "../m3u8.c"
 void gf_xml_dump_string(FILE* file, const char *before, const char *str, const char *after) {}
 
+/*************************************/
+
+static Bool hls_output_contains(FILE *output, const char *marker)
+{
+	char text[4096];
+	size_t bytes_read;
+
+	fflush(output);
+	rewind(output);
+	memset(text, 0, sizeof(text));
+	bytes_read = fread(text, 1, sizeof(text) - 1, output);
+	text[bytes_read] = 0;
+	return strstr(text, marker) ? GF_TRUE : GF_FALSE;
+}
+
+unittest(mpd_hls_scte35_dynamic_rewrite_preserves_cues)
+{
+	GF_MPD mpd = {0};
+	GF_MPD_Period period = {0};
+	GF_MPD_AdaptationSet adaptation_set = {0};
+	GF_MPD_Representation representation = {0};
+	GF_DASH_SegmentContext segment = {0};
+	GF_MPD_EventStream event_stream = {0};
+	GF_MPD_EventStreamEntry event_entry = {0};
+	FILE *first_playlist;
+	FILE *second_playlist;
+	GF_Err error;
+
+	period.event_streams = gf_list_new();
+	event_stream.entries = gf_list_new();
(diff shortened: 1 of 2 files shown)


src/media_tools/unittests/ut_mpd.c
```diff
@@ -5,6 +5,86 @@
...
+static Bool hls_output_contains(FILE *output, const char *marker)
+{
+	char text[4096];
+	size_t bytes_read;
+
+	fflush(output);
+	rewind(output);
+	memset(text, 0, sizeof(text));
+	bytes_read = fread(text, 1, sizeof(text) - 1, output);
+	text[bytes_read] = 0;
```
rbouqueau, line 19: This is already 0 whatever. We also have a preference for `char text[4096] = {0};` and removing the `memset` call.

rbouqueau:
One minor comment, LGTM otherwise. Thanks for all these PRs, much appreciated.

PR #3915, opened 2026-09-17
fix(rmtws): correct JS_GetArrayBuffer size type 

PierreDillard:
`js_sys_rmt_client_send` stored the ArrayBuffer size in a `u32` but passed
  its address to `JS_GetArrayBuffer`, which expects `size_t*` (8 bytes on
  64-bit builds). This overflowed the 4-byte variable and corrupted stack
  memory on  binary RMT message.

  Fix: use `size_t` for `bufsize`.


jeanlf:
thanks!

